PR #53, opened 2019-09-16
include config in languages definition

lindensc:
Hi, i have modified package.json to include configuration directive in objectscript-class definition.
Also i have created a copy of the language-definition-class.jsonc with the name class-language-definition.json.

This fixes the issue that if you toggle line comment in objectscript-class was using "#;" instead of "//".

Best Regards

Timo


daimor:
Did it programmatically, so, M code and class lines, will have different comments

PR #152, opened 2020-06-18
Add more studio actions

ty-d:
Addresses the following studio extension VSCode gaps:
- MenuItem save flags are not supported
- No differentiation between different menus - Added the menu name to the right of the menu item
- Studio actions produce JSON output - Added an option to enable the output, but it is disabled by default
- For prompts, AfterUserAction supplies the user's text in Answer instead of Msg (with Answer indicating yes/no/cancel); also, there's no yes/no/cancel, it's just text or cancel
- Actions can't open CSP pages (which should include authentication, similar to Atelier)
- Actions can't open documents
- Actions can't modify document text
- ObjectScript explorer needs to support context menus - A context menu was added to the explorer that displays the available context actions
- No support for disabling menu items - Disabled items no longer appear in the list of actions
- Items marked read-only appear to be editable and even saved, but are not actually.
- Items that are not read-only, but are editable, should fire an "other" action with Type 1, Name 0 when edit is attempted (and support UserAction/AfterUserAction).
- Should fire Type 1, Name 5 ("changed namespace") when connection is established.

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,6 +157,10 @@
           "when": "view == ObjectScriptExplorer && viewItem =~ /^dataNode:/"
         },
         {
+          "command": "vscode-objectscript.studio.contextActions",
+          "when": "view == ObjectScriptExplorer && viewItem =~ /^dataNode:/"
+        },
+        {
           "command": "vscode-objectscript.explorer.otherNamespace",
           "when": "view == ObjectScriptExplorer && viewItem =~ /^serverNode((?!:extra:).)*$/",
           "group": "inline"
@@ -410,6 +414,11 @@
         "category": "ObjectScript",
         "command": "vscode-objectscript.studio.actions",
         "title": "Studio actions"
+      },
+      {
+        "category": "ObjectScript",
+        "command": "vscode-objectscript.studio.contextActions",
+        "title": "Studio context actions"
       }
     ],
     "keybindings": [
--- a/package.json
+++ b/package.json
@@ -624,6 +624,11 @@
           "type": "boolean",
           "default": true,
           "description": "Show `Debug this method` action for ClassMethods"
+        },
+        "objectscript.studioActionDebugOutput": {
+          "type": "boolean",
+          "default": false,
+          "description": "Output the action that VSCode should perform as requested by the server in JSON format."
         }
       }
     },
--- a/package.json
+++ b/package.json
@@ -413,12 +413,12 @@
       {
         "category": "ObjectScript",
         "command": "vscode-objectscript.studio.actions",
-        "title": "Studio actions"
+        "title": "Studio Actions"
       },
       {
         "category": "ObjectScript",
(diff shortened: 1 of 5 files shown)


timleavitt:
@daimor , for some context, this is trying to get the %Studio.Extension.* capabilities in vscode-objectscript up to par with Atelier (at least). This is important for internal adoption at InterSystems as well as for many other use cases within the community.

daimor:
yeah, sure, thanks, I'll have a look

gjsjohnmurray:
A good tip from @daimor is that the CI process means we can easily grab a VSIX to test this out with even before it gets merged.

Start here:
image

Then go here:
image

Download the zip, unpack it, then install the package.vsix file within.

Nice!

package.json
```diff
@@ -410,6 +414,11 @@
         "category": "ObjectScript",
         "command": "vscode-objectscript.studio.actions",
         "title": "Studio actions"
```
gjsjohnmurray, line 416: To conform to VS Code style this title should be `Studio Actions...`

src/commands/studio.ts
```diff
@@ -148,3 +349,38 @@ export async function mainMenu(uri: vscode.Uri) {
...
+    case OtherStudioAction.DeletedDocument:
+      label = "Deleted Document";
+    case OtherStudioAction.OpenedDocument:
+      label = "Opened Document";
+    case OtherStudioAction.ClosedDocument:
+      label = "Closed Document";
+    case OtherStudioAction.ConnectedToNewNamespace:
+      label = "Changed Namespace";
+    case OtherStudioAction.FirstTimeDocumentSave:
+      label = "Saved Document to Server for the First Time"
```
gjsjohnmurray, line 383: For consistency add trailing semicolon

src/commands/studio.ts
```diff
@@ -121,19 +270,71 @@ class StudioActions {
...
+
+    // Save the current document
+    if(bitString.charAt(0) === "1") {
+      await saveAndCompile(vscode.window.activeTextEditor.document);
+    }
+
+    // Save all documents
+    if(bitString.charAt(2) === "1") {
+      for(const document of vscode.workspace.textDocuments) {
+        await saveAndCompile(document)
```
gjsjohnmurray, line 334: For consistency add trailing semicolon

src/commands/studio.ts
```diff
@@ -36,41 +71,144 @@ class StudioActions {
...
+              offset = +method.split("+")[1];
+              method = method.split("+")[0];
+            }
+          }
+
+          const splitClassname = classname.split(".");
+          const filetype = splitClassname[splitClassname.length - 1];
+          const isCorrectMethod = (text: string) => (filetype === "cls")
+            ? text.match("Method " + method)
+            : text.startsWith(method)
```
gjsjohnmurray, line 159: For consistency add trailing semicolon

PR #164, opened 2020-06-22
fix-163 Studio Context Actions command errored if no file open

### LINKED ISSUE #163: Command 'ObjectScript: Studio Context Actions...' resulted in an error (Cannot read property 'document' of undefined)
This command arrived with PR #152. It appears unconditionally in the Command Palette and causes an error if invoked when no document is open.

I will supply a PR.

gjsjohnmurray:
PR to fix #163 

@ty-d, @timleavitt before getting this PR merged I'd like to understand the different roles of the `vscode-objectscript.studio.actions` and `vscode-objectscript.studio.contextActions` commands.

The XData of %Studio.SourceControl.Base defines two menus:

- %SourceMenu
- %SourceContext

I subclassed %Studio.SourceControl.Base and added an extra option at the beginning of each menu in order to differentiate them:

```
XData Menu
{
<MenuBase>
<Menu Name="%SourceMenu" Type="0">
<MenuItem Name="SourceMenu"/>
<MenuItem Name="%CheckOut"/>
<MenuItem Name="%UndoCheckout"/>
<MenuItem Name="%CheckIn" Save="100"/>
<MenuItem Name="%GetLatest"/>
<MenuItem Name="%AddToSourceControl" Save="100"/>
</Menu>
<Menu Name="%SourceContext" Type="1">
<MenuItem Name="SourceContext"/>
<MenuItem Name="%CheckOut"/>
<MenuItem Name="%UndoCheckout"/>
<MenuItem Name="%CheckIn" Save="100"/>
<MenuItem Name="%GetLatest"/>
<MenuItem Name="%AddToSourceControl" Save="100"/>
</Menu>
</MenuBase>
}
```

Studio uses the former for its menubar Source Control menu,

image

and the latter for the Source Control item on its context menus (i.e. right-click) on open documents 

image

and items in trees in the Workspace panel:

image

In VS Code the `objectscript.studio.actions` command supplies the quickpick we get from VS Code context menus on isfs-scheme files in VS Code's native Explorer **and** on the editor of an open isfs-scheme file. Its caption is `Studio Actions...`

My test class results in this quickpick:

image

This lists the active entries from %SourceMenu followed by any active members from %SourceContext that haven't previously been added.

The `objectscript.studio.contextActions` command appears on the context menu of the ObjectScript Explorer view, Its caption is 'Studio Context Actions...` and populates my quickpick as follows:

image

This lists active members from %SourceContext only.

From the above analysis I have several questions:

1. Why does an isfs item in ObjectScript Explorer have a different context-menu-initiated Studio Context Actions quickpick from the Studio Actions one on other context menus (editor/context and explorer/context)?

2. Why does one of those quickpicks (Studio Actions) include options from both XData Menus?

3. Was it intentional that the `ObjectScript: Studio Context Actions..." command always appeared in the Command Palette?

Re question 3, a change in this PR suppresses that command completely from the palette. But without understanding the thinking behind these two commands I'm not certain that's the correct thing to have done.


ty-d:
I tried to make the "Studio Context Actions" menu in the explorer be similar to Studio's and only populate the context actions. Since there is no equivalent of a top menu in VSCode, the "Studio Actions" command was originally made to show both types of menus, so I did not change the behavior of that quickpick. It was not intentional to always show the context actions command.

If it is okay to display both types of menus from the explorer context menu, then there would be no need to have the "Studio Context Actions" command (it's not used anywhere else).

package.json
```diff
@@ -128,11 +128,19 @@
           "when": "false"
         },
         {
-          "command": "vscode-objectscript.studio.actions",
+          "command": "vscode-objectscript.serverCommands.sourceControl",
```
timleavitt, line 131: @gjsjohnmurray I don't like "sourceControl" here - maybe call it serverCommands.studioExtension? (The "studio extension" framework is more general than just source control.)

gjsjohnmurray: @timleavitt with this PR I have implemented separate menus for the Source Control commands and the rest, to replicate how Studio behaves. This command brings up the %SourceMenu menu items, which Studio displays under a menu named "Source Control". The command `vscode-objectscript.serverCommands.other` brings up the items from all other menus for which context=0.

`vscode-objectscript.serverCommands.contextSourceControl` shows the %SourceContext menu's items, and `vscode-objectscript.serverCommands.contextOther` shows all the other ones for which context=1

If you'd like to see this in action you can always get the vsix artifact that our CI pipeline creates whenever I push stuff to this PR:

image

Install this in your VS Code and give it a try. I'll be glad to more of your feedback.

timleavitt: @gjsjohnmurray thanks - that makes more sense now. I'll try it out when I get a chance.

PR #186, opened 2020-07-06
Add new file prompt to explorer

ty-d:
Adds a "New..." option in the explorer that opens a prompt where the user can fill in the name of a new file. The file is then saved to the server and opened in an editor.

### DIFF
--- a/src/commands/newFile.ts
+++ b/src/commands/newFile.ts
@@ -1,0 +1,132 @@
+import * as vscode from "vscode";
+import { AtelierAPI } from "../api/index";
+import { DocumentContentProvider } from "../providers/DocumentContentProvider";
+import { NodeBase } from "./../explorer/models/nodeBase";
+import { RootNode } from "./../explorer/models/rootNode";
+import { PackageNode } from "../explorer/models/packageNode";
+import { ClassNode } from "../explorer/models/classesNode";
+import { RoutineNode } from "../explorer/models/routineNode";
+import { fireOtherStudioAction, OtherStudioAction } from "./studio";
+
+export async function createNewFile(node: NodeBase) {
+  try {
+    const name = await vscode.window.showInputBox(getInputBoxOptions(node));
+    // Check if user cancelled
+    if (!name) {
+      return;
+    }
+
+    const api = new AtelierAPI();
+    const response = await api.putDoc(name, { enc: false, content: getContent(name) }, true);
+
+    vscode.commands.executeCommand("vscode-objectscript.explorer.refresh");
+    const uri = DocumentContentProvider.getUri(response.result.name, node.workspaceFolder, node.namespace);
+    if (response.result.ext && response.result.ext[0] && response.result.ext[1]) {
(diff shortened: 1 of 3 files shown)


daimor:
I’m not sure that we need this.
Explorer view not for editing, if need to create new class or routine it should be done from files explorer.
Explorer is just a viewer.

ty-d:
@timleavitt @daimor This feature was requested for those using a server-side workflow.

timleavitt:
@daimor I think this has significant value when working in a server-centric mode. If you're primarily editing files in the "server-side editing" setting, with server-side source control (e.g., your local filesystem doesn't really matter) why should you need to switch contexts and create a new file on disk in order to add a new class/routine?

gjsjohnmurray:
> 
> 
> @timleavitt @daimor This feature was requested for those using a server-side workflow.

I tend to agree with @daimor. It's unclear to my why those using a server-side workflow are using ObjectScript Explorer at all. Doesn't an isfs-type root folder in the standard VS Code Explorer view meet their needs? This already has a new-file command button, and a new-file entry on the context menu.

src/commands/newFile.ts
```diff
@@ -0,0 +1,132 @@
...
+      contentArray = ["Class " + name, "{", "", "}"];
+      break;
+    case "MAC":
+      contentArray = ["ROUTINE " + name, ""];
+      break;
+    case "INC":
+      contentArray = ["ROUTINE " + name + " [Type=INC]", ""];
+      break;
+    case "CSP":
+      contentArray = [
```
timleavitt, line 111: Note: the file is not necessarily HTML. What happens if you create js/css/etc.?

ty-d: That is a switch on the file extension, so js and css files are empty when created.

timleavitt: Oh. Duh. Thanks!

daimor:
For server-side editing have to be used isfs feature, where this ability already available.

src/commands/newFile.ts
rajrsingh: @timleavitt are we agreed that the ObjectScript Explorer view will stay read-only? I also like the simplicity of keeping all editing to the normal Explorer view.

timleavitt: @rajrsingh yes; I actually wasn't aware of isfs. (Or rather, I'd heard of it and incorrectly equated it to the ObjectScript explorer.) Do we have doc on setting that up somewhere? I did see it in https://community.intersystems.com/post/your-opinion-matters-intersystems-development-environments#comment-82141

rajrsingh: @timleavitt the best doc we have on it right now is https://github.com/intersystems-community/vscode-objectscript/wiki/Server-side-editing but of course our internal team will be working hard this month on that. They could probably benefit from your usage experience.

rajrsingh:
I believe resolution of this discussion is to NOT merge this PR. Will close by end of week if no other comments are made.

PR #188, opened 2020-07-07
Make ServerManager extension as optional, do not prevent this extension working.

### LINKED ISSUE #187: Strict extension dependency
Instead of strictly say that this extension depends on `intersystems-community.servermanager` we should offer it as an option. At the moment this extension will not work until dependency has not solved, but it could work without it anyway.

daimor:
resolves #187

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,9 +64,6 @@
     "onDebugInitialConfigurations"
   ],
   "main": "./dist/extension",
-  "extensionDependencies": [
-    "intersystems-community.servermanager"
-  ],
   "contributes": {
     "menus": {
       "commandPalette": [
@@ -712,6 +709,11 @@
           "type": "boolean",
           "default": false,
           "description": "Output the action that VSCode should perform as requested by the server in JSON format."
+        },
+        "objectscript.ignoreInstallServerManager": {
+          "type": "boolean",
+          "default": false,
+          "markdownDescription": "Do not offer to install `#intersystems-community.servermanager#` extension"
         }
       }
     },
--- a/package.json
+++ b/package.json
@@ -708,12 +708,12 @@
         "objectscript.studioActionDebugOutput": {
           "type": "boolean",
           "default": false,
-          "description": "Output the action that VSCode should perform as requested by the server in JSON format."
+          "description": "Output the action that VS Code should perform as requested by the server in JSON format."
         },
         "objectscript.ignoreInstallServerManager": {
           "type": "boolean",
           "default": false,
-          "markdownDescription": "Do not offer to install `#intersystems-community.servermanager#` extension"
+          "markdownDescription": "Do not offer to install the [intersystems-community.servermanager](https://marketplace.visualstudio.com/items?itemName=intersystems-community.servermanager) helper extension."
         }
       }
     },
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -263,8 +263,7 @@ async function serverManager(): Promise<void> {
   }
   return vscode.window
     .showInformationMessage(
-      `Extension InterSystems® Server Manager have not installed,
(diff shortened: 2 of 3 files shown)


src/extension.ts
```diff
@@ -253,6 +253,40 @@ export const checkConnection = (clearCookies = false): void => {
...
+      "Install",
+      "Skip",
+      "Ignore"
+    )
+    .then(async (action) => {
+      switch (action) {
+        case "Install":
+          await vscode.commands.executeCommand("workbench.extensions.installExtension", extId);
+        // eslint-disable-next-line no-fallthrough
+        case "View":
```
gjsjohnmurray, line 277: showInformationMessage doesn't offer "View" option

src/extension.ts
```diff
@@ -253,6 +253,40 @@ export const checkConnection = (clearCookies = false): void => {
...
+          await vscode.commands.executeCommand("workbench.extensions.search", `@tag:"intersystems"`);
```
gjsjohnmurray, line 278: Why not simply search for extId?

src/extension.ts
```diff
@@ -253,6 +253,40 @@ export const checkConnection = (clearCookies = false): void => {
...
+  const ignore =
+    config("ignoreInstallServerManager") ||
+    vscode.workspace.getConfiguration("intersystems.servers").get("/ignore", false);
+  if (ignore || vscode.extensions.getExtension(extId)) {
+    return;
+  }
+  return vscode.window
+    .showInformationMessage(
+      `Extension InterSystems® Server Manager have not installed,
+      it's recommended to install it.`,
```
gjsjohnmurray, line 267:
```suggestion
      "The InterSystems® Server Manager extension is recommended to help you define connections."
```

package.json
```diff
@@ -712,6 +709,11 @@
           "type": "boolean",
           "default": false,
           "description": "Output the action that VSCode should perform as requested by the server in JSON format."
```
gjsjohnmurray, line 711:
```suggestion
          "description": "Output the action that VS Code should perform as requested by the server in JSON format."
```

package.json
```diff
@@ -712,6 +709,11 @@
...
+        },
+        "objectscript.ignoreInstallServerManager": {
+          "type": "boolean",
+          "default": false,
+          "markdownDescription": "Do not offer to install `#intersystems-community.servermanager#` extension"
```
gjsjohnmurray, line 716: The syntax you used is only for hyperlinking to other settings.
```suggestion
          "markdownDescription": "Do not offer to install the [intersystems-community.servermanager](https://marketplace.visualstudio.com/items?itemName=intersystems-community.servermanager) helper extension."
```

daimor:
This extension still can work without servermanager, but with extensiondependency, it does not work at all, until servermanager installed. And it’s not silently, it asks user for installation approval.

src/extension.ts
daimor: search for extId just will show only one extension on the left panel, so, I decided to show any intersystems tagged extensions, but open particular extension in full view.

src/extension.ts
daimor: My thoughts were, do not offer too many buttons, and was not sure if it's a good idea to install it from information message, not from extension page

gjsjohnmurray: In which case why is `case "View":` here?

src/extension.ts
gjsjohnmurray: Since there's no direct path to `case "View":` the user picks "Install", the servermanager gets installed, then you show them all intersystems-tagged extensions in the list on the left, and the servermanager one displays in detail on the right. Is that what you mean? Seems a bit unexpected, but I can live with it.

PR #189, opened 2020-07-07
Show system items in ObjectScript Explorer view

### LINKED ISSUE #165: "System" classes in non %SYS namespaces are not visible
Any classes starting with `%` aren't visible to VS Code. Many developers create % classes so there should be an option to view them similar to how generated assets are handled. 

This happens in `rootNode.ts` where it says:
`    const systemFiles = this.namespace === "%SYS" ? "1" : "0";`

The user should be able to toggle between showing "system files" or not, no matter the namespace.

daimor:
closes #165

### DIFF
--- a/package.json
+++ b/package.json
@@ -196,6 +196,14 @@
         {
           "command": "vscode-objectscript.explorer.hideGenerated",
           "when": "view == ObjectScriptExplorer && viewItem =~ /^serverNode.*:generated:/"
+        },
+        {
+          "command": "vscode-objectscript.explorer.showSystem",
+          "when": "view == ObjectScriptExplorer && viewItem =~ /^serverNode((?!:(%SYS|system):).)*$/"
+        },
+        {
+          "command": "vscode-objectscript.explorer.hideSystem",
+          "when": "view == ObjectScriptExplorer && viewItem =~ /^serverNode.*:system:/"
         }
       ],
       "editor/context": [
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -401,6 +409,16 @@
         "title": "Hide Generated",
         "category": "ObjectScript"
       },
+      {
+        "command": "vscode-objectscript.explorer.showSystem",
+        "title": "Show System",
```
rajrsingh, line 414: Since %Z classes aren't system classes, I suggest this menu item name be "Show % Classes"

reference: https://docs.intersystems.com/irislatest/csp/docbook/Doc.View.cls?KEY=GORIENT_appx_identifiers

package.json
```diff
@@ -401,6 +409,16 @@
...
+        "category": "ObjectScript"
+      },
+      {
+        "command": "vscode-objectscript.explorer.hideSystem",
+        "title": "Hide System",
```
rajrsingh, line 419: Since %Z classes aren't system classes, I suggest this menu item name be "Hide % Classes"

reference: https://docs.intersystems.com/irislatest/csp/docbook/Doc.View.cls?KEY=GORIENT_appx_identifiers

rajrsingh:
> LGTM. Just to clarify, this feature only applies to ObjectScript Explorer, right? If so, maybe make that clear in the PR title.

Good point @gjsjohnmurray. @daimor would it make sense to have the same capability with the ISFS view?

package.json
daimor: %ZEN also %Z

daimor:
isfs has much more capabilities, it can redefine namespace, use filters and types of files

rajrsingh:
> isfs has much more capabilities, it can redefine namespace, use filters and types of files
> https://github.com/intersystems-community/vscode-objectscript/wiki/Server-side-editing

So isfs already has the capability to do this via manual filters.

@daimor I don't know why the build is failing...

daimor:
it can't test extension due to dependency, the extension can't run if no dependency resolved. will be good after merge #188

Issue #167, opened 2020-06-26
Working With Namespaces


zkrowiak:
Editing .cls files from distinct namespaces is ambiguous as they all populate the same space in the src/ file.  It would be nice if  files from separate namespaces could be placed in separate subdirectories.  

That would also account for the edge case of there being two .cls files with the same name that represent different classes on different namespaces.

daimor:
But how you configured so, to have different namespaces?

zkrowiak:
I clicked 'view another namespace' and exported a class from that namespace. On looking again I see that it exports from that namespace but compiles and adds it to the direct namespace. Is there a way to edit multiple namespaces currently?

daimor:
Yep, manual export from any namespace in the explorer view, just exports it to the local folder, the same, no matter about namespace. 
You can deal with multiple namespaces, or even servers, in the same view with multi-root feature, where each root folder, will have its own server configuration, and for sure, it's own src folder.
Look at [this example](https://github.com/caretdev/vscode-multi-connection/)

zkrowiak:
ok thank you. That makes sense.  Perhaps the warning that comes up when you try to export from a separate namespace inform users that the resulting export will not work the way they are expecting.
![image](https://user-images.githubusercontent.com/54324676/85884289-641d7280-b7b0-11ea-8bc4-751b29d500c6.png)

daimor:
That's just a warning, you may suggest a new text for it, so it would be much more clear to its reader.

zkrowiak:
How about:  "You are about to export items from another namespace.  Edits to these files on your system will be compiled in the primary namespace for your root, not the namespace from which exported them.  Would you like to continue?

PR #207, opened 2020-07-22
Export from another namespace, message updated

daimor:
Resolves: #167

### DIFF
--- a/src/commands/export.ts
+++ b/src/commands/export.ts
@@ -192,7 +192,7 @@ export async function exportExplorerItem(nodes: NodeBase[]): Promise<any> {
     const answer = await vscode.window.showWarningMessage(
       `
 You are about to export items from another namespace.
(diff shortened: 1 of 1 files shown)


src/commands/export.ts
```diff
@@ -191,7 +191,8 @@ export async function exportExplorerItem(nodes: NodeBase[]): Promise<any> {
   if (origNamespace !== node.namespace) {
     const answer = await vscode.window.showWarningMessage(
       `
-You are going to export items from another namespace.
+You are about to export items from another namespace.
+Edits to these files on your system will be compiled in the primary namespace for your root, not the namespace from which exported them.
```
gjsjohnmurray, line 195:
```suggestion
Edits to these files in your local workspace will be compiled in the primary namespace of your workspace root, not the namespace from which you originally exported them.
```

PR #228, opened 2020-07-28
fix #162 pass folder name to ObjectScript Explorer root node

### LINKED ISSUE #162: Unable to create multi-environment workspace
Hi Dmitriy,

Here's the layout in VS Code's Explorer:

image

The code-workspace file from the Workspace top-level directory:

image

Note that I've tried both manually creating the file and adding directories to the workspace using the VS Code menus. The only difference I can see in the result is that the json name value appears in the explorer tree rather than the actual directory name.

Heres's a .vscode\settings.json file from one of the directories referenced in the code-workspace file shown above:

image

The oddest symptom is that when I change just one of the settings.json files, all directories change to those value in the OBJECTSCRIPT:EXPLORER view (only the file I've actually edited and saved has the changes in it, though):

image

Also, another weird issue is that if I select a directory in the OBJECTSCRIPT:EXPLORER view and choose "View another namespace," it opens the namespace for that directory but duplicates all of the existing directories including the one that just had it's namespace changed.

Before:

image

Select namespace (FTP):

image

After:

image

Thanks for looking into this.

PS.

Just in case it matters:

Version: 1.46.1 (user setup)
Commit: cd9ea6488829f560dc949a8b2fb789f3cdc05f5d
Date: 2020-06-17T21:13:20.174Z
Electron: 7.3.1
Chrome: 78.0.3904.130
Node.js: 12.8.1
V8: 7.8.279.23-electron.0
OS: Windows_NT x64 10.0.18362

gjsjohnmurray:
This PR fixes #162

### DIFF
--- a/src/explorer/explorer.ts
+++ b/src/explorer/explorer.ts
@@ -73,11 +73,14 @@ export class ObjectScriptExplorerProvider implements vscode.TreeDataProvider<Nod
         const conn: any = config("conn", workspaceFolder.name);
         if (conn.active && conn.ns) {
           const extra4Workspace = this._showExtra4Workspace[workspaceFolder.name] || [];
-          node = new WorkspaceNode(workspaceFolder.name, this._onDidChangeTreeData, {});
+          node = new WorkspaceNode(workspaceFolder.name, this._onDidChangeTreeData, {
+            workspaceFolder: workspaceFolder.name,
+          });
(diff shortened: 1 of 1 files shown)


gjsjohnmurray:
Fix verified at https://github.com/intersystems-community/vscode-objectscript/issues/162#issuecomment-665255631

PR #238, opened 2020-07-30
fix #237 Don't open wrong local document from ObjectScript Explorer

### LINKED ISSUE #237: ObjectScript Explorer's view of another namespace incorrectly opens a local file if the name matches
1. Set up a single-root workspace with an `objectscript.conn` that connects to namespace A on a server, in which there is a routine named Test1.int
2. Use ObjectScript Explorer (OE), expand the Routines folder and export Test1.int to the local workspace folder.
3. In namespace B on the same server, create a Test1.int routine that isn't the same as the one in A.
4. Use OE to "View Another Namespace", and pick B.
5. Expand the Routines folder of B, find Test1.int and click on it to open the file.
6. 🐛 The local copy of Test1.int (exported from A) opens. It should have opened the server-side copy from namespace B.

gjsjohnmurray:
This fixes #237.

It also adds visual differentiation between files in OE that will open from the server and those that will open from the local workspace.

image

Problem indicators will display for the latter.

### DIFF
--- a/src/explorer/models/cspFileNode.ts
+++ b/src/explorer/models/cspFileNode.ts
@@ -20,6 +20,7 @@ export class CSPFileNode extends NodeBase {
         command: "vscode-objectscript.explorer.openCSPFile",
         title: "Open File",
       },
+      resourceUri: isLocalFile ? itemUri : undefined,
       contextValue: CSPFileNode.contextValue,
       label: `${displayName}`,
       tooltip: isLocalFile ? undefined : this.fullName,
--- a/src/explorer/models/cspFileNode.ts
+++ b/src/explorer/models/cspFileNode.ts
@@ -10,16 +10,19 @@ export class CSPFileNode extends NodeBase {
 
   public getTreeItem(): vscode.TreeItem {
     const displayName: string = this.label;
+    const itemUri = DocumentContentProvider.getUri(this.fullName, this.workspaceFolder, this.namespace);
+    const isLocalFile = itemUri.scheme === "file";
 
     return {
       collapsibleState: vscode.TreeItemCollapsibleState.None,
       command: {
-        arguments: [DocumentContentProvider.getUri(this.fullName, this.workspaceFolder, this.namespace)],
+        arguments: [itemUri],
         command: "vscode-objectscript.explorer.openCSPFile",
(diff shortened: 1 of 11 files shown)


src/explorer/models/cspFileNode.ts
```diff
@@ -10,16 +10,19 @@ export class CSPFileNode extends NodeBase {
...
 
     return {
       collapsibleState: vscode.TreeItemCollapsibleState.None,
       command: {
-        arguments: [DocumentContentProvider.getUri(this.fullName, this.workspaceFolder, this.namespace)],
+        arguments: [itemUri],
         command: "vscode-objectscript.explorer.openCSPFile",
         title: "Open File",
       },
       contextValue: CSPFileNode.contextValue,
```
daimor, line 23: resourceUri missed?

src/explorer/models/workspaceNode.ts
```diff
@@ -33,19 +33,59 @@ export class WorkspaceNode extends NodeBase {
...
 
-    node = new RootNode("Routines", "", "dataRootNode:routinesRootNode", "RTN", this.options);
+    node = new RootNode(
+      "Routines",
+      "",
+      "dataRootNode:routinesRootNode",
+      "RTN",
+      this.options,
+      false
+      //new vscode.ThemeIcon("note")
```
daimor, line 54: Why commented?

gjsjohnmurray: Because at the moment there seems to be a tree bug which messes up alignment if some items (i.e. local files) have icons and others (remote) don't. I only see this bug when the containing folder has an icon.

I plan to report this on the VSCode project. Assuming it gets fixed we can uncomment the lines that add icons to our folders.

src/explorer/models/cspFileNode.ts
gjsjohnmurray: I don't understand what you mean.

daimor: any other node class has `resourceUri` field here as well.
```suggestion
      resourceUri: isLocalFile ? itemUri : undefined,
      contextValue: CSPFileNode.contextValue,
```

Issue #241, opened 2020-07-31
Check if AfterUserAction method of server-side support class is called in the right circumstances


gjsjohnmurray:
Documentation of AfterUserAction method includes the following sentence:

> For the cases where Studio did not perform any interaction this method is not called.

I didn't spot any logic in studio.ts to make its AfterUserAction call equivalently conditional.

Investigate this, e.g. using a dummy server-side class that logs its calls. Check what we do matches what Studio does.

gjsjohnmurray:
@ty-d are you willing to take this task?

ty-d:
Yes, I can look into this.

PR #246, opened 2020-08-03
Open CSP pages and improve AfterUserAction

ty-d:
Open Studio Extension CSP pages without a progress bar in the lower right.

### DIFF
--- a/src/commands/studio.ts
+++ b/src/commands/studio.ts
@@ -277,11 +277,14 @@ class StudioActions {
               (actionToProcess) =>
                 actionToProcess &&
                 !afterUserAction &&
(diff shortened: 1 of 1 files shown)


gjsjohnmurray:
Looks fine to me, but for future reference we'd like an Issue to be created first so there's somewhere detailing/discussing what the PR is fixing/improving.

ty-d:
Sorry, I'll open an issue next time. CSP pages were not opening since the action was not being passed to processUserAction.

I also added an if statement to check that there is an answer being received from processUserAction (indicating interaction with the user) before running AfterUserAction to address issue #241.

Issue #261, opened 2020-08-11
Customised exporting structure


daimor:
@daimor Ok thanks, as a test I have set my export as follows:

```
"objectscript.export": {
        "folder": {
            "*./abstract/*.cls": "abstract/abstract"
        },
        "addCategory": true
     }
```

This exports my file to ```[object Object]/cls/abstract```

Where I am hoping it to export to ```cls/abstract/abstract```

Is there any reason it is adding this ```[object Object]``` onto the front of the export?

_Originally posted by @howardbailey in https://github.com/intersystems-community/vscode-objectscript/issues/75#issuecomment-671850609_

daimor:
@howardbailey

Ok, `[object object]` comes from folder, which is supposed to be as string only as a root folder for any of your sources in the repo. 
While your pattern should be in `addCategory`. I suppose that you have classes like `somepackage.abstract.someclass.cls` and you would like this.
```
    "objectscript.export": {
      "folder": "src",
      "addCategory": {
        ".*\\.Abstract\\..*\\.cls": "cls/Abstract",
        "cls": "cls",
        "mac": "mac"
      }
    },
```
And how it exports
```
export "User.Abstract.Person.cls" as "/Users/daimor/temp/src/cls/Abstract/User/Abstract/Person.cls" - Success 
Exported items: 1
export "User.Person.cls" as "/Users/daimor/temp/src/cls/User/Person.cls" - Success 
```
You should keep in mind, if you would still need separate folders for each type, e.g. cls, mac and so on, you will have to define it as I did above.

Or this
```
    "objectscript.export": {
      "folder": "src",
      "addCategory": {
        "Abstract\\..*\\.cls": "cls/Abstract",
        "cls": "cls",
        "mac": "mac"
      }
    },
```
to get 
```
export "Abstract.Test.cls" as "/Users/daimor/temp/src/cls/Abstract/Abstract/Test.cls" - Success 
```

howardbailey:
Ok great thanks, let me try that.

daimor:
I've found the issue in order of checking patterns, it would not allow more then one pattern to check, I'm going to fix it. As well as an issue with type of `folder` is not string, or if it's just empty.

howardbailey:
Ok great thanks.

I've tried more or less what you have suggested above.

```
"objectscript.export": {
     "folder": "CLS",
     "addCategory": {
        ".*\\.abstract\\..*\\.cls": "CLS/abstract/abstract",
     }
[...]

PR #262, opened 2020-08-11
fixes in configuration export.addCategory

daimor:
resolves: #261


gjsjohnmurray:
@daimor if you add a brief description of what this fixes, then merge it, there'll be something I can add to the 0.8.9 changelog.

daimor:
Resolves error when something else instead of string was used in the `folder` value.
if `addCategory` contains multiple patterns, should check all of them, in given order, previously of if only the first pattern checked if was defined pattern by type or for any.

PR #265, opened 2020-08-12
fix #254 detect doubleclick on ObjectScript Explorer items

### LINKED ISSUE #254: Files opened from ObjectScript Explorer use VS Code preview mode unless this is disabled
On the main VS Code Explorer the default is that a single click opens a file in preview mode. Its tab label is _italicized_ and the tab is re-iused when another file is opened. Opening with a double-click prevents the file's tab from being re-used. A preview-mode tab can be persisted by double-clicking the tab.

Our ObjectScript Explorer doesn't follow this pattern. Files get opened in preview mode even when double-clicked.

gjsjohnmurray:
This PR fixes #254


gjsjohnmurray:
@daimor if you're able to verify this from the VSIX, or at least approve it after code review, it can go in our next release, 0.8.9. I'd like to ship that soon because it will fix some ISFS-related bugs that got introduced in 0.8.8.

PR #268, opened 2020-08-13
fix #256 reload file explorer after connection

### LINKED ISSUE #256: Failed Authentication on "Expanded" Server
When closing a workspace/exiting VS Code, the Explorer view is cached. However, when the workspace is re-opened, a failed authentication on a server whose view was previously "expanded" to show classes/routines will show the arrow indicator as though it is expanded, but with an empty list.

A successful authentication does not refresh the list automatically, nor does clicking the expanded indicator (arrow) next to the server entry. Clicking the Refresh icon **does** display the expanded view, however.

The expanded indicator is a cosmetic issue, but I feel that a successful authentication to a server should trigger a refresh for that server's view.

gjsjohnmurray:
This PR fixes #256

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -196,6 +196,7 @@ export function checkConnection(clearCookies = false, uri?: vscode.Uri): void {
         outputChannel.appendLine(
           `Something is wrong with your docker-compose connection settings, or your service is not running.`
         );
+        outputChannel.show(true);
         panel.text = `${packageJson.displayName} - ERROR`;
         return;
       }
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -292,6 +296,9 @@ export function checkConnection(clearCookies = false, uri?: vscode.Uri): void {
     })
     .finally(() => {
       explorerProvider.refresh();
+      if (uri.scheme === FILESYSTEM_SCHEMA || uri.scheme === FILESYSTEM_READONLY_SCHEMA) {
```
daimor, line 299:
```suggestion
      if (schemes.includes(uri.scheme)) {
```

gjsjohnmurray:
@daimor I have applied your suggested change. Please approve again.

Issue #126, opened 2020-04-28
Search option on the server side


grongierisc:
Is it possible to have search option on the server side ?

API :
localhost:57772/api/atelier/v2/:namespace/action/search

daimor:
Unfortunately, not [everything is finished](https://github.com/microsoft/vscode/issues?q=is%3Aissue+is%3Aopen+label%3Asearch+label%3Aapi-finalization) on VSCode side, to make it easier in use.
It is already implemented here in some way, but not not publicly available, yet.

gjsjohnmurray:
@grongierisc if you want to try server-side searching, here's how:

1. Install VS Code Insiders
2. In it, install the beta VSIX from https://github.com/intersystems-community/vscode-objectscript/releases/tag/v0.8.9-beta.9
3. Follow [these instructions](https://github.com/intersystems-community/vscode-objectscript/pull/270#issuecomment-674080534) to make Insiders able to give our extension the proposed APIs.
4. Use a workspace with an isfs:// type folder.

gjsjohnmurray:
Starting with 0.9.3-beta.2 you will now see this in the ObjectScript output channel as confirmation you've configured things right to get proposed APIs.
```
intersystems-community.vscode-objectscript version 0.9.3-beta.2 activating with proposed APIs available.
```

copperschnack:
Thanks @gjsjohnmurray - Unfortunately, I see no output on `ObjectScript `output channel. Neither when just using `argv.json` setting, nor when additionally providing the directive via command line. (I do not run VS Code insiders with elevated privileges).

![image](https://user-images.githubusercontent.com/14861913/93080658-5cbe8480-f68e-11ea-8d5b-6e3c726b9108.png)

daimor:
you should use `intersystems-community.vscode-objectscript`  instead of `daimor.vscode-objectscript`

copperschnack:
@daimor - well seen! I was using the previous id from https://github.com/intersystems-community/vscode-objectscript/pull/270#issuecomment-674080534. Search is working now :-)

Thank you!

gjsjohnmurray:
@copperschnack it seems you don't actually need to be using Insiders to make this work. I have updated my previous comment above.

copperschnack:
@gjsjohnmurray - very right and even better!

copperschnack:
[...]

PR #270, opened 2020-08-14
fix #269 Server-side search of isfs folder using proposed API failed when using Server Manager connection

### LINKED ISSUE #269: Server-side search of isfs folder using proposed API failed when using Server Manager connection
Despite their package.json containing the expected `"enableProposedApi": true` setting, the FileSearchProvider and TextSearchProvider implementations don't seem to activate.

I suspect this is caused by the build procedure not adding vscode.proposed.d.ts

I will submit a PR.

gjsjohnmurray:
This PR fixes #269

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,7 +54,6 @@ jobs:
       with:
         node-version: 12.x
     - run: npm install
-    - run: npx vscode-dts dev
     - name: lint
       if: runner.os == 'Linux'
       run: npm run lint
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,6 +54,7 @@ jobs:
       with:
         node-version: 12.x
     - run: npm install
+    - run: npx vscode-dts dev
     - name: lint
       if: runner.os == 'Linux'
       run: npm run lint
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -172,7 +172,7 @@ export class AtelierAPI {
       } = getResolvedConnectionSpec(serverName, config("intersystems.servers", workspaceFolderName).get(serverName));
       this._config = {
         active: this.externalServer || conn.active,
-        apiVersion: 1,
+        apiVersion: workspaceState.get(this.configName + ":apiVersion", DEFAULT_API_VERSION),
         https: scheme === "https",
         ns: namespace || conn.ns,
         host,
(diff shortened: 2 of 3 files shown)


daimor:
image
Now it doing the same twice

gjsjohnmurray:
Yes I spotted that. Am investigating. Hence this is only a draft PR.

gjsjohnmurray:
The original problem (proposed APIs not available) turned out to have been caused by me having forgotten that it's necessary to add the `--enable-proposed-api daimor.vscode-objectscript' argument when launching vscode-insider, otherwise our extension behaves as though its package.json specified `"enableProposedApi": false` even though the manually-installed VSIX (beta or snapshot) has this set to true.

Simpler than having to launch the Insiders executable from the command-line with this extra argument is to do this:
1. In Command Palette choose `Preferences: Configure Runtime Arguments`
2. In the argv.json file that opens, add this line:
`	"enable-proposed-api": ["daimor.vscode-objectscript"]`
3. Restart Insiders

Now you can use VS Code's Search (Ctrl/Cmd+Shift+F) to run a server-side search on your isfs folder(s), as long as the server supports v2 of the Atelier API.

copperschnack:
> Now you can use VS Code's Search (Ctrl/Cmd+Shift+F) to run a server-side search on your isfs folder(s), as long as the server supports v2 of the Atelier API.

I cannot get the search to work in my environment: Querying for `classmethod` retrieves no results, where obviously it should.

* VS Code `1.50.0-insider (user setup)`
* Plugin `vscode-objectscript v0.9.3-beta.1`
* IRIS `2020.1` with `/api/atelier/v2` available and other server side content displayed correctly in VS Code workspace

image

daimor:
it's maybe not obvious, but you cannot find all classmethods in any classes this way. It may find `$classmethod` in the code, or if you have it as text somewhere, but not just any classmethod. Could you try to compare it with Studio Search?

daimor:
I would suggest trying to search for something else, not for keywords.

copperschnack:
OK, thanks for the hint. I retried with a different search and added a comparision to Studio in a screenshot below. Our IRIS instance is an In-Place upgraded Caché-Instance. Just mentioning it, in case there might be caviots there.

image

daimor:
Could you show if the Output in VSCode shows anything? It supposed to be exactly the same as in Studio.

copperschnack:
There ist no output regarding searching. Compilation instead does appear like in Studio, see below:

`Compilation started on 09/14/2020 11:07:24 with qualifiers 'cuk', using up to 4 worker jobs`
`Compiling class tmp.SearchTest`
`Compiling routine tmp.SearchTest.1`
`Compilation finished successfully in 0.026s.`

daimor:
Are you sure, that you started VSCode correctly?
Sorry, issue maybe just in renamed extension id, you have to use `intersystems-community.vscode-objectscript` instead of `daimor.vscode-objectscript`

gjsjohnmurray:
I have updated the instructions above to reflect the new extension id.

Also, see https://github.com/intersystems-community/vscode-objectscript/issues/126#issuecomment-691966405 for a new indicator that you have set things up correctly. @copperschnack I suggest continuing this discussion on #126 rather than on this closed PR.

PR #306, opened 2020-08-28
Launching CSP pages with server manager settings

### LINKED ISSUE #305: Legacy studio extension CSP page launch not supported when only using server manager
When using the server manager (rather than fully specified objectscript.conn settings), launching a CSP page from a legacy Studio extension doesn't work. It looks like this was missed when adding server manager support.

(I've identified a fix and will open a PR.)

timleavitt:
This PR fixes #305


daimor (approved):

gjsjohnmurray (approved):
LGTM

PR #314, opened 2020-09-03
fix #302: manage use of vscode.window.showInputBox

### LINKED ISSUE #302: Several issues with server manager and multiple server connections
Use case:
In the user-level settings.json file I have several entries in intersystems.servers. In a workspace-level settings file I have one isfs folder definition using each of them.

If I don't specify the username for each in intersystems.servers:
I get prompted for username/password on one of them (the last one in the list, it happens), and after that connections succeed. Connections continually fail for the other with:
Authorization error: Check your credentials in Settings, and that you have sufficient privileges on the /api/atelier web application on host:57772[NAMESPACE]
Presumably it's assumed I want to act as UnknownUser.

If I specify the username for each in intersystems.servers, I see this repeatedly:
Authorization error: Check your credentials in Settings, and that you have sufficient privileges on the /api/atelier web application on host:57772[NAMESPACE]
I am prompted for a password on each of the instances, but each time the prompt replaces the current prompt, removing any input I may have typed before I may have had time to submit it.

It would be great to have behavior around asking for credentials tightened a bit. I'm a fan of the secure storage of these credentials in the Server Manager, but as it stands this is a major barrier to connecting to multiple servers easily.

(@rajrsingh indicated that this is a vscode-objectscript issue rather than intersystems-servermanager)

timleavitt:
Fixes #302 by only showing one input box at a time, deduplicating per server

### DIFF
--- a/src/utils/inputBoxManager.ts
+++ b/src/utils/inputBoxManager.ts
@@ -2,19 +2,41 @@ import vscode = require("vscode");
 
 import { InputBoxOptions } from "vscode";
 
-// Used in situations where multiple input boxes may be displayed to ensure that we don't stomp on
-// the box that's already shown.
+// Interface used internally to track queued input boxes
+interface InputBoxSpec {
+  key?: string;
+  options: InputBoxOptions;
+  callback: (answer: string) => void;
+}
+
+/**
+ * Used in situations where multiple input boxes may be displayed in rapid succession
+ * to allow the user to respond to them one at a time.
+ */
 export class InputBoxManager {
+  /** Set to true if an input box is currently shown */
   private static shown = false;
+
+  /** Array of input boxes to show */
   private static specs = new Array<InputBoxSpec>();
+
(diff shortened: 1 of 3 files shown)


src/utils/inputBoxManager.ts
```diff
@@ -0,0 +1,59 @@
...
+        InputBoxManager.next();
+      },
+      (ignored) => {
+        InputBoxManager.next();
+      }
+    );
+  }
+}
+
+interface InputBoxSpec {
```
gjsjohnmurray, line 55: Perhaps move this to near the top, above the class definition.

gjsjohnmurray (changes requested):
Nice solution, and a cursory test of the vsix artifact the PR built worked well for me on the workspace I'd been seeing the problem on.

Perhaps the InputBoxManager class would benefit from a few inline comments to convey what it's doing.

timleavitt:
@gjsjohnmurray I've added comments as requested.

gjsjohnmurray (approved):
Thanks. Looks great!

Issue #330, opened 2020-09-11
Provide easier way to link a local folder to a server


gjsjohnmurray:
Similar to #321 (which focused on server-side editing situations) it may be possible to add helptext and a button onto the empty ObjectScript Explorer view that appears when I open a folder that doesn't yet have a .vscode/settings.json file with a suitable `objectscript.conn` definition, and there is no user-level definition to fall back to.

I plan to investigate this idea.

PR #338, opened 2020-09-16
Add command to link local folder to server (#330)

gjsjohnmurray:
This PR implements #330 

With one or more local folders open but lacking `objectscript.conn` settings, the ObjectScript view displays a button that runs the new `connectFolderToServerNamespace` command, and a link to `hideExplorerForWorkspace` (also new).

image

 Together with a new `showExplorerForWorkspace` these commands are also available on the palette.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,8 @@
       },
       {
         "view": "ObjectScriptExplorer",
-        "contents": "Connect a workspace folder to an [InterSystems server](https://intersystems-community.github.io/vscode-objectscript/serverside/) if you want to export or import ObjectScript code.\n[Choose Server and Namespace](command:vscode-objectscript.connectFolderToServerNamespace)\nOr [hide this ObjectScript view](command:vscode-objectscript.hideExplorerForWorkspace) if it is not needed for the current workspace."
+        "contents": "Connect a workspace folder to an [InterSystems server](https://intersystems-community.github.io/vscode-objectscript/serverside/) if you want to export or import ObjectScript code.\n[Choose Server and Namespace](command:vscode-objectscript.connectFolderToServerNamespace)\nOr [hide this ObjectScript view](command:vscode-objectscript.hideExplorerForWorkspace) if it is not needed for the current workspace.",
+        "when": "vscode-objectscript.explorerRoots == 0"
       }
     ],
     "menus": {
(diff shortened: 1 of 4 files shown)


gjsjohnmurray:
@rajrsingh if you have a few minutes, maybe try this out. You can get a VSIX containing it from inside the zip downloadable here:
image

In VS Code, open a folder that doesn't yet have a .vscode\settings.json file defining `objectscript.conn`. Go to the ObjectScript view. If you see a tree instead of the new button you've probably defined `objectscript.conn` at the User level. Remove that from your settings.json, or comment it out. Then go back to the ObjectScript view.

rajrsingh:
@gjsjohnmurray if I run VS Code and open a new window (no folder selected), this doesn't work. Were you expecting to handle that use case?

gjsjohnmurray:
@rajrsingh with no folder open I don't even get the ObjectScript button (ISC icon) in the Activity Bar. Do you?

I've just pushed a change to the welcome view of VS Code Explorer to handle this case:

image

What do you think?

gjsjohnmurray:
FYI https://github.com/microsoft/vscode/issues/106487 is about the way the Clone Repository button comes late to the party.

rajrsingh:
> @rajrsingh with no folder open I don't even get the ObjectScript button (ISC icon) in the Activity Bar. Do you?
> 
I don't get the activity bar button but I do get the "Choose Server and Namespace" button in the Explorer pane. That's the workflow I was trying.

gjsjohnmurray:
Please try the latest VSIX from the Checks tab of this PR.

rajrsingh:
I thought it wasn't saving settings I entered via the palette, but now I see it's buried in `/Library/Application\ Support/Code\ -\ Insiders/User/settings.json`.

I think the flow is a bit confusing but it works as long as you don't make a mistake, so I'm going to approve it as a first cut that could use refinement.

gjsjohnmurray:
@rajrsingh Thanks for commenting. I have now requested a review from you. Once you approve it I will merge.

Re finding the buried JSON file, an easier way is using `Preferences: Open Settings (JSON)`. Also reachable from this button top right in Settings Editor:

image

rajrsingh:
@gjsjohnmurray that button is a great tip we should have in docs!!

rajrsingh (approved):

gjsjohnmurray:
@rajrsingh please update the docs to use the new and simpler pathways for getting started, both for local editing (typically git controllled) and server-side editing. Now the server selector includes a way to define a new server it should be possible for both use cases to be productive without writing any JSON.

PR #358, opened 2020-10-01
Added warning message to Jump to Tag and Offset

### LINKED ISSUE #356: Jump to Tag + Offset: fail more descriptively on non-int/mac files
Jump to Tag + Offset only works on int and mac files. If I run it on a cls file it fails without any indication. I'd prefer a warning message that tells me the file types it supports.

pravinb1:
This PR fixes #356.

### DIFF
--- a/package.json
+++ b/package.json
@@ -203,6 +203,10 @@
         {
           "command": "vscode-objectscript.showExplorerForWorkspace",
           "when": "!config.objectscript.showExplorer"
+        },
+        {
+          "command": "vscode-objectscript.jumpToTagAndOffset",
+          "when": "editorLangId == objectscript"
         }
       ],
       "view/title": [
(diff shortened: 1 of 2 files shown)


gjsjohnmurray:
@pravinb1 An alternative / additional strategy is to add a few lines to package.json so this command only appears in the Command Palette when the languageId is `objectscript`. By default only .int and .mac files use this, and it's also this languageId that makes the status bar item show the current position (from which a click is a shortcut to the Jump to Tag and Offset command. 

```json
    "menus": {
      "commandPalette": [
        {
          "command": "vscode-objectscript.jumpToTagAndOffset",
          "when": "editorLangId == objectscript"
        },
```

pravinb1:
Thanks for the tip @gjsjohnmurray, I've added the condition in package.json.

gjsjohnmurray (approved):
LGTM

gjsjohnmurray:
@daimor if you're happy with this I think it's ready to be merged. @pravinb1 is that correct?

daimor (approved):

PR #360, opened 2020-10-02
initial support for working together with language-server

daimor:
This PR fixes #

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -99,6 +99,8 @@ const packageJson = vscode.extensions.getExtension(extensionId).packageJSON;
 const extensionVersion = packageJson.version;
 const aiKey = packageJson.aiKey;
 
+const _onDidChangeConnection = new vscode.EventEmitter<void>();
+
 export const config = (setting?: string, workspaceFolderName?: string): vscode.WorkspaceConfiguration | any => {
   workspaceFolderName = workspaceFolderName || currentWorkspaceFolder();
   if (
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -525,10 +535,50 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
     outputChannel.show(true);
   }
 
+  const languageServerExt = await languageServer();
+  if (!languageServerExt) {
```
gjsjohnmurray, line 539: I suggest you add an `outputChannel.appendLine` call in this block to notify user that the language server wasn't found.

gjsjohnmurray (commented):
With both extension installed and active, if I disable the LS extension it requires a reload. Things then work as expected, with the ObjectScript extension taking over all the roles. But if I subsequently enable LS no reload is required and we get contributions from both extensions (e.g. double hovers).

Here's an idea for a solution:

- During `activate` gather the Disposables returned from each of the `onXXX` and  `registerXXX` calls that you make only when LS isn't active.
- Use the [`extensions.odDidChange` event](https://code.visualstudio.com/updates/v1_31#_extensions-change-event) to detect when LS gets activated. When this happens, dispose of all the Disposables you gathered during activation.

daimor:
> With both extension installed and active, if I disable the LS extension it requires a reload. Things then work as expected, with the ObjectScript extension taking over all the roles. But if I subsequently enable LS no reload is required and we get contributions from both extensions (e.g. double hovers).
> 
> Here's an idea for a solution:
> 
> * During `activate` gather the Disposables returned from each of the `onXXX` and  `registerXXX` calls that you make only when LS isn't active.
> * Use the [`extensions.odDidChange` event](https://code.visualstudio.com/updates/v1_31#_extensions-change-event) to detect when LS gets activated. When this happens, dispose of all the Disposables you gathered during activation.

Done

gjsjohnmurray (approved):
LGTM and I will merge it immediately.

PR #362, opened 2020-10-02
Fix #361, support opening "other" doc types via isfs

### LINKED ISSUE #361: Unable to open "other" document types under isfs
Currently isfs filters to the following extensions:
cls|mac|int|inc|dfi|bpl

There are other document types as well (.LUT, .HL7, and other user-level customizations via subclasses of %Studio.AbstractDocument).

Using an isfs uri with filter=* will show these in the folder, but they currently can't be opened.

These documents need to be openable, and also to support source control context commands. (This is a partial blocker for adoption of VSCode in HealthShare development within InterSystems.)

timleavitt:
Also enables them for source control actions even if they're XML files (e.g., DFI and some other custom document types used in Interoperability and HS)

This PR fixes #361

### DIFF
--- a/package.json
+++ b/package.json
@@ -286,12 +286,12 @@
         },
         {
           "command": "vscode-objectscript.serverCommands.contextSourceControl",
-          "when": "resourceScheme == isfs && editorLangId =~ /^objectscript/ && vscode-objectscript.connectActive",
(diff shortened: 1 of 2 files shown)


daimor (approved):

gjsjohnmurray (approved):
LGTM

PR #366, opened 2020-10-07
Fix extension reload

### LINKED ISSUE #259: Server side source control with isfs not reloading if revert done in perforce
I have a class that I reverted using the perforce client, but when I open the class using the isfs connection it is not reloaded from disk but instead shows the changed class prior to the revert.  If I go to studio and open the class it is reloaded from disk, and then switching back to VSCode it reloads the class in the editor.

I would expect that when I revert the class in the perforce client that opening the class or switching to the class in the editor would reload from disk (assuming of course it hasn't been changed).

0.8.9-beta.5

timleavitt:
This PR does not fix #259 but does address https://github.com/intersystems-community/vscode-objectscript/issues/259#issuecomment-682088028 and some related issues my team has observed.

Use case:
We have a Source Control > Undo Check Out menu item for our Perforce-based extension. This will return reload=true. The problem was twofold - we weren't checking/acting on the "reload" flag from AfterUserAction, and when we did, we were detecting the revert as an edit and blocking it unless the user checked the file out again (obviously not correct).

There isn't really a clean way to detect this case, so we now track URIs for which we've requested a reload and ignore the next "attempted edit" on them. While not the neatest, the solution is at least localized to the legacy studio command set of functionality, and I don't see alternatives.

### DIFF
--- a/src/commands/studio.ts
+++ b/src/commands/studio.ts
@@ -59,6 +59,8 @@ function getOtherStudioActionLabel(action: OtherStudioAction): string {
   return label;
 }
 
+const suppressEditListenerMap = new Map<string, boolean>();
+
 class StudioActions {
   private uri: vscode.Uri;
   private api: AtelierAPI;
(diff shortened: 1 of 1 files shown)


daimor (approved):

gjsjohnmurray (approved):
A sensible pragmatic solution.

PR #384, opened 2020-10-15
Support server-side (isfs) folder settings, snippets, debug configs etc

### LINKED ISSUE #383: Support folder-specific settings when using the server-side editing paradigm (isfs)
VS Code's settings model allows a folder-specific \.vscode\settings.json file to contain settings that take precedence when operating under that workspace root folder.

If a workspace root folder uses isfs to operate directly in a namespace on a server, it is not currently possible for that namespace to store and serve up a file named \.vscode\settings.json

I will submit a PR that overcomes this limitation.

The \.vscode subfolder of a workspace root is also where folder-specific code snippets and folder-specific debug configurations. The enhancement will mean these can be shared among developers who connect to a common server namespace.

gjsjohnmurray:
This PR resolves #383

To provide the server-side storage, define a web application named `/_vscode` like this:

image

(alter the Physical Path value to suit your platform and install folder)

An isfs-type workspace root folder that connects to a namespace on this server can now write and read folder-specific settings, e.g.

image

Folder-specific snippets file can be created via `Preferences: Configure User Snippets`

image

To edit the server-side namespace-specific files (all namespaces) directly through VS Code, add an isfs-type root folder with the following uri:

```
isfs://servername/_vscode?ns=%SYS&csp
```

For a single namespace (e.g USER):

```
isfs://servername/_vscode/USER?ns=%SYS&csp
```

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -496,4 +496,12 @@ export class AtelierAPI {
     };
     return this.request(1, "GET", `%SYS/jobs`, null, params);
   }
+
+  // v1+
+  public getCSPApps(detail = false): Promise<Atelier.Response> {
+    const params = {
+      detail: detail ? 1 : 0,
+    };
+    return this.request(1, "GET", `%SYS/cspapps/${this.ns || ""}`, null, params);
+  }
 }
--- a/src/commands/compile.ts
+++ b/src/commands/compile.ts
@@ -251,18 +251,22 @@ export async function importAndCompile(askFlags = false, document?: vscode.TextD
       // console.error(error);
       throw error;
     })
-    .then(() => compile([file], flags));
+    .then(() => {
+      if (!file.fileName.startsWith("\\.vscode\\")) {
+        compile([file], flags);
+      }
+    });
 }
 
 // Compiles all files types in the namespace
-export async function namespaceCompile(askFLags = false): Promise<any> {
+export async function namespaceCompile(askFlags = false): Promise<any> {
   const api = new AtelierAPI();
   const fileTypes = ["*.CLS", "*.MAC", "*.INC", "*.BAS"];
   if (!config("conn").active) {
     throw new Error(`No Active Connection`);
   }
   const defaultFlags = config().compileFlags;
-  const flags = askFLags ? await compileFlags() : defaultFlags;
+  const flags = askFlags ? await compileFlags() : defaultFlags;
   if (flags === undefined) {
     // User cancelled
     return;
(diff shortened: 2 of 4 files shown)


gjsjohnmurray:
@timleavitt members of your team might be interested in trying this new capability.

timleavitt:
@gjsjohnmurray I think you're right - thank you!

gjsjohnmurray:
Bonus benefit of this work is that a csp-type isfs root folder in a workspace can handle situations where the namespace is home to multiple webapps and you want an overview (i.e. not one root per app). For example, here's what currently happens:

image

And again with this PR:

image

@daimor any chance of you reviewing this today? I'd like to merge it for the next release.

daimor:
In %SYS Namespace with csp=1 flag I see lots of inc files
image

And In Ensemble enabled namespace, in addition to inc files, I can see dtl, and x12 files
image

gjsjohnmurray:
@daimor ready for you to look at again

daimor:
There is one more issue, it shows more CSP applications than should. It should skip CSP application with no physical path, such as set to work with Dispatch Class.

Can you add this code to the end of api/index.ts
```
  // v1+
  public getCSPApps(): Promise<Atelier.Response> {
    return this.request(1, "GET", `%SYS/cspapps/${this.ns || ""}`);
  }
```

and use the output of this, to get available CSP Applications?
result of this query is very simple, just an array of path's, as you can see, it does not contain any `/api/` application
```
            "/csp/broker",
            "/csp/documatic",
            "/csp/sys",
            "/csp/sys/exp",
            "/csp/sys/mgr",
            "/csp/sys/op",
            "/csp/sys/sec",
            "/isc/studio/rules",
            "/isc/studio/templates",
            "/isc/studio/usertemplates"
```

daimor (changes requested):
use getCSPApps function to get the list of correct CSP applications

gjsjohnmurray:
Will that work for a user without access to %SYS ?

daimor:
Yes, sure, it in fact does not have any checks for it.
%SYS in API URLs does not mean that it needs access to it.

daimor:
```
  // v1+
  public getCSPApps(detail = false): Promise<Atelier.Response> {
    const params = {
      detail: detail ? 1 : 0,
    };
    return this.request(1, "GET", `%SYS/cspapps/${this.ns || ""}`, null, params);
  }
```
If add details, and pass true, it will return it with the default flag
```
            {
                "name": "/csp/user",
                "default": true
            }
```

daimor (approved):

jstetson10:
@gjsjohnmurray : I've got some questions, likely not of general interest. Can you send me your email address? Mine is jstetson@intersystems.com

PR #407, opened 2020-10-21
Different fixes in CI

daimor:
- Resolves the issue with draft releases
- Publish prebuild package 
- changes in artifacts

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,17 +5,17 @@ on:
     branches:
       - master
     paths-ignore:
-      - 'docs/**'
-      - '.vscode/**'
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (approved):
Looks good, though the proof will come when we do the next release.

gjsjohnmurray:
Thanks @daimor. If you're willing to make the same changes in https://github.com/intersystems-community/intersystems-servermanager then I can easily do a two-stage release of that to publish the tweaked icon.

gjsjohnmurray:
The equivalent CI changes on Server Manager were verified by releasing 1.0.1 there. Thanks @daimor

PR #413, opened 2020-10-22
Ability to add file mapping

### LINKED ISSUE #412: Change file names during export
Needs the ability to transform file names during export.
for instance any % classes, should use underscore instead of the percent sign
%ZPM.PackageManager.cls have to be stored on the disk as _ZPM/PackageManager.cls

daimor:
This PR fixes #412

### DIFF
--- a/package.json
+++ b/package.json
@@ -746,6 +746,37 @@
           ],
           "default": false
         },
+        "objectscript.export.map": {
+          "markdownDescription": "Map file names before export with regexp pattern as a key, and replacement as a value\n (e.g. ```JSON\n {\n  \"%(.*)\": \"_$1\"\n}\n``` for % class or routines replaced to be with underscore instead).",
+          "type": "object",
+          "default": {},
+          "examples": [
+            {
+              "%(.*)": "_$1"
+            }
+          ],
+          "patternProperties": {
+            ".*": {
+              "type": "string"
+            }
+          },
+          "properties": {
+            "%(.*)": {
+              "description": "Catch any % items, with value as `_$1` will replace `%` to `_`",
+              "type": "string",
+              "examples": [
+                "_$1"
+              ]
+            },
+            "(.*)": {
+              "description": "Complete pattern, use `$1` to get matched value",
+              "type": "string",
+              "examples": [
+                "$1"
+              ]
+            }
+          }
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -746,6 +746,37 @@
           ],
           "default": false
         },
+        "objectscript.export.map": {
+          "markdownDescription": "Map file names before export with regexp pattern as a key, and replacement as a value\n (e.g. ```JSON\n {\n  \"%(.*)\": \"_$1\"\n}\n``` for % class or routines replaced to be with underscore instead).",
```
gjsjohnmurray, line 750:
```suggestion
          "markdownDescription": "Map file names before export, with regexp pattern as a key and replacement as a value (e.g. `{  \"%(.*)\": \"_$1\" }` to make % classes or routines use underscore prefix instead).",
```
Improve wording, and remove Markdown elements not supported in this property.

package.json
```diff
@@ -746,6 +746,37 @@
...
+            }
+          ],
+          "patternProperties": {
+            ".*": {
+              "type": "string"
+            }
+          },
+          "properties": {
+            "%(.*)": {
+              "description": "Catch any % items, with value as `_$1` will replace `%` to `_`",
```
gjsjohnmurray, line 765:
```suggestion
              "description": "Catch any % items. Set value to `_$1` to replace `%` prefix with `_`",
```

package.json
```diff
@@ -746,6 +746,37 @@
...
+              "type": "string",
+              "examples": [
+                "_$1"
+              ]
+            },
+            "(.*)": {
+              "description": "Complete pattern, use `$1` to get matched value",
```
gjsjohnmurray, line 772:
```suggestion
              "description": "Match everything. Set value to `$1` to leave unchanged",
```

gjsjohnmurray (changes requested):

gjsjohnmurray:
@daimor I just committed a change to your branch that improves defaulting and adds validation. Please review.

Also, why is it necessary to avoid filenames starting with the % character?

daimor:
It's not necessary, it's the ability for compatibility with existing source code conventions

gjsjohnmurray (approved):

PR #430, opened 2020-10-24
ci fix for enableProposedApi

### LINKED ISSUE #429: v1.0.2 failed to publish to Marketplace
See https://github.com/intersystems-community/vscode-objectscript/runs/1300135166?check_suite_focus=true

image

I don't understand why this happened when I created the v1.0.2 release but not with the previous v1.0.1

It seems that some step in the script didn't toggle `enableProposedApi` in settings.json when it should have.

@daimor please investigate.

daimor:
This PR fixes #429

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -153,9 +153,9 @@ jobs:
           npm install
           git push
           echo ::set-output name=ovsx::`cat .name`
-      - uses: actions/download-artifact@v2
-        with:
-          name: ${{ steps.set-version.outputs.name }}.vsix
+      - name: Build package
+        run: |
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (approved):
Thanks for this. Once you have merged it will you be able to get CI to publish 1.0.2? If not then I can do a 1.0.3 release.

daimor:
Yes

daimor:
the trick did not happen, it just fails because the version already updated in package.json
so, better to create new release

gjsjohnmurray:
The 1.0.3 release published to Marketplace but not to Open VSX Registry.

image

Issue #409, opened 2020-10-21
Suggested doc improvements


gjsjohnmurray:
A task list to collect/track improvements to what is published at https://intersystems-community.github.io/vscode-objectscript/

(checkmarks updated to reflect #441 changes)
### Installation.md
- [x] Refresh Marketplace screenshot so it shows icons and more recent version & publisher information.
- [x] Once https://github.com/intersystems-community/vscode-objectscript-pack publishes, suggest install of the extension pack instead of the individual extensions. A search for 'intersystems' returns more than just the extensions needed.
- [x] ObjectScript Explorer icon only appears on the activity bar when a folder or workspace has been opened.

### workspace.md
- [x] It is not true that a .code-workspace file is always in the root folder.

### Configuration.md
- [x] An 'Add Server' user story that begins from the '+' button of a Server Manager quickpick is significantly simpler for newcomers than the current one which requires opening a JSON file and editing that. In which case, relegate the details of the JSON structure to a later section about editing an existing connection spec.
- [x] PR #408 reinstated references to three `InterSystems Server Manager: Test ...` commands that no longer exist. Remove these references.
- [x] The need to add an `objectscript.conn` to JSON settings can also be avoided. Open the folder you want to put client-side files in. Click on the ObjectScript Explorer icon on the activity bar. Use the 'Choose Server and Namespace' button. Pick a server (or add a new one). Pick the namespace on the server. That's it. The required `objectscript.conn` will get created in the .vscode\settings.json file of your folder.
- [x] The info under 'Configuring Export from Server' is really for advanced users in the situation where the default export behaviour doesn't meet their needs. Don't alarm newcomers with it.

### ServerSide.md
- [x] Fix the `=======` string that appears above the step-by-step list.
- [x] In step 4 add a note that a new server definition can be added using the '+' button.
- [x] Expand step 7 to cover how this part of the process now has four options (two each for code and webapps).
- [x] 'Web Application (CSP) Files' section is advanced info now that the above step 7 supports adding of a webapp-type folder to the workspace.

jstetson10:
I'll need some input regarding what the current status of Web Application (CSP) Files is and how they should be handled in the documentation.

PR #441, opened 2020-10-27
Docs v1 updates

isc-rsingh:
Various updates including some fixes for #409


docs/Configuration.md
```diff
@@ -2,107 +2,102 @@
...
 ```
-The components of this server definition are:
+
+The components of this configuration are:
 
 - **ns** - namespace to use on the server
 - **server** - server name as specified in the server configuration
-- **active** - specifies whether the connection is active.
-
-## Configuring Export from Server
```
daimor, line 79: Why this section is deleted?

isc-rsingh: @daimor we are moving some content around to group it better, so that may be why, but @jstetson10 may be able to comment more.

gjsjohnmurray: @daimor, in #409 I wrote re this section:

> The info under 'Configuring Export from Server' is really for advanced users in the situation where the default export behaviour doesn't meet their needs. Don't alarm newcomers with it.

gjsjohnmurray (approved):
Seems to contain a good number of improvements. Worth publishing, even though more is likely to be done to it.

docs/Configuration.md
```diff
@@ -2,107 +2,102 @@
...
-		"webServer": {
-			"scheme": "http",
-			"host": "localhost",
-			"port": 52774,
-		},
-	"username": "_SYSTEM",
-	},
-},
-```
+Once you have entered these values, the server definition is stored in your user settings. At that point another prompt appears, asking for a password to store in the system Keychain. Enter the password for the username supplied earlier to complete the process.
```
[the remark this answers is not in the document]
isc-rsingh: Good tip. But don't bale out by hitting escape because it won't be created, right? Bale out by pressing return.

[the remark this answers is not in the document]
gjsjohnmurray: No, provided you don't press Escape at the http/https prompt or before, your new server definition will have been filed. Escape at the password entry field will bale out from storing a password, but the new server definition will remain. I'm planning to add an info notification at the point where the new server definition has been saved, in the same way as there's one when a password has been stored or removed from the keychain.

PR #475, opened 2020-11-04
CHANGELOG and README for 1.0.5

gjsjohnmurray:
@daimor we need to publish 1.0.5 before LS 1.0.5 so at to avoid FoldingRangeProvider conflict.

cc @isc-bsaviano

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Change Log
 
+## [1.0.5] 5-Nov-2020
+- Defer to Language Server 1.0.5+ for folding range information (#473)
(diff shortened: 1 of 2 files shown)


isc-bsaviano:
@gjsjohnmurray @daimor We aren't planning on releasing LS version 1.0.5 this week so you don't need to rush this out if there are other things pending.

daimor (approved):

daimor (approved):

gjsjohnmurray:
@isc-bsaviano FYI, 1.0.5 has been published so we're ready for LS 1.0.5 whenever it comes.

PR #491, opened 2020-11-09
fix #490 make Documatic launcher class-specific

### LINKED ISSUE #490: Are there any plans to allow VS Code to launch Documatic for the current class?
This is supported in Studio and Atelier

gjsjohnmurray:
This PR fixes #490 

The server action menu, accessed by clicking on the connection's panel in the status bar, now targets its Open Class Documentation onto the currently open class if any.

Side-effects of this change:
- the ${classname} substitution no longer includes a redundant ".cls" suffix. This is a breaking change but seems low-risk because this feature doesn't appear to have been documented yet.
- a new ${classnameEncoded} substitution has been added, suitable for use in a URL querystring.

### DIFF
--- a/src/commands/serverActions.ts
+++ b/src/commands/serverActions.ts
@@ -59,10 +59,15 @@ export async function serverActions(): Promise<void> {
         return;
       });
   }
+  const file = currentFile();
+  const classname = file && file.name.toLowerCase().endsWith(".cls") ? file.name.slice(0, -4) : "";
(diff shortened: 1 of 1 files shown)


daimor (approved):

isc-solon:
...and thanks to this fix, I am now aware of the "server action menu!" I was right-clicking the connection instead of clicking. Is this menu documented anywhere?

PR #497, opened 2020-11-10
Readme cleanup

mvhenderson:
* Use consistent whitespace in readme code-blocks
* Minor language tweak in proposed-api section to make it clear VS Code needs to shutdown for changes to take effect (not just reload window)


gjsjohnmurray (approved):
Thanks for the PR

PR #499, opened 2020-11-10
correct work status bar with docker

### LINKED ISSUE #498: Click on Status panel does not show Server actions
when using docker-compose setting in objectscript.conn
If would like to refresh connection from menu, but it does not available, and when I click nothing happens

daimor:
This PR fixes #498

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -32,6 +32,8 @@ export interface ConnectionSettings {
   ns: string;
   username: string;
   password: string;
+  docker: boolean;
+  dockerService?: string;
 }
 
 export class AtelierAPI {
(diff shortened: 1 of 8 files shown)


gjsjohnmurray (commented):
Are the changes in TextSearchProvider.ts connected to the problem of status bar panel with Docker? Or are they for something else?

daimor:
In TextSearchProvider just to ignore any errors. If error happens it just cleans any search results

gjsjohnmurray:
> 
> 
> Are the changes in TextSearchProvider.ts connected to the problem of status bar panel with Docker? Or are they for something else?

OK, though I'd prefer a PR not to contain multiple unrelated fixes. Anyway, now that #496 has merged you'll need to resolve conflicts here.

package.json
```diff
@@ -46,6 +46,7 @@
   },
   "enableProposedApi": true,
   "activationEvents": [
+    "*",
```
gjsjohnmurray, line 49: Why do you want to activate the extension unconditionally now? It will mean that even if I launch VS Code to view or edit a .txt file our extension will start. Probably not much of an overhead, but it should still be justifiable.

And if we **do** adopt this I suggest all the other entries should be removed because they serve no purpose.

daimor: It looks like it's only one way to add a panel to the status bar, for some empty folder.

gjsjohnmurray: I prefer the current behaviour. Opening an empty folder doesn't add ObjectScript panel to the status bar but **does** add the ObjectScript item to the activity bar. Clicking that icon activates the extension, so the status bar panel now appears. The empty ObjectScript view also gives the user some help about what to do next.

gjsjohnmurray (approved):

PR #511, opened 2020-11-13
fix #510 respect pathPrefix on server action URLs

### LINKED ISSUE #510: URLs in Server Actions menu should respect a `pathPrefix` value
When the connection comes from an `intersystems.servers` entry that includes a `pathPrefix` value this should be incorporated into URLs on the Server Actions menu (e.g. Open Management Portal)

gjsjohnmurray:
This PR fixes #510

It also adds pathPrefix info to the "Connected as" tooltip of the status bar panel when pathPrefix is set.

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -137,9 +137,9 @@ export class AtelierAPI {
   }
 
   public xdebugUrl(): string {
-    const { host, https, port, apiVersion } = this.config;
+    const { host, https, port, apiVersion, pathPrefix } = this.config;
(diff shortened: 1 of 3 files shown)


daimor:
Could you update `xdebugUrl` function as well (in `api/index.ts`), which is used for debugging?

daimor (approved):

gjsjohnmurray:
> 
> 
> Could you update `xdebugUrl` function as well (in `api/index.ts`), which is used for debugging?

@daimor I've done this and it's ready for your review again.

PR #526, opened 2020-12-13
Add snippets from Projection and XData

isc-kiyer:
This PR fixes #480

PR #529, opened 2020-12-15
Enable "View other" for custom Studio documents

isc-kiyer:
"View other" should work not just for COS classes/routines. Custom Studio documents could be XML or in some other format and have output from "View other". Hence, when in a folder in a workspace that has the ObjectScript extension enabled, whatever document is in context should allow you to use "View other" because it is not possible to predict all possible formats of custom Studio documents.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
         },
         {
           "command": "vscode-objectscript.viewOthers",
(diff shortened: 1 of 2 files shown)


daimor (approved):

gjsjohnmurray (commented):
This PR has included your changes from PR #526. Please avoid doing this in future by starting your new branch from master.

PR #533, opened 2020-12-17
Fix leading / issue that prevents custom Studio documents from being saved on the server

isc-kiyer:
Custom Studio documents will never have / in their name (which is only applicable to CSP files). Hence, if the "name" variable in currentFile() has a leading /, it should be stripped so that the document name is passed correctly to the Atelier API when updating on the document on the server.


gjsjohnmurray (approved):
LGTM

daimor (approved):

PR #534, opened 2020-12-17
Support searching for all Studio document types when using Cmd + T.

isc-kiyer:
Current implementation of symbol search (Cmd/Ctrl + T) is to only search for classes, routines and methods. Expand this default to also search include files.
Added a setting to toggle whether ALL studio document types should be searched for rather than just the defaults.
Default is still to only search classes, routines, methods and include files (added).
Can toggle on searching for all Studio documents. Default is to NOT search (as it can be expensive if many CSP files present).
Removed byClasses because it can be handled by StudioOpenDIalog.

### DIFF
--- a/package.json
+++ b/package.json
@@ -833,6 +833,11 @@
           "default": "cuk",
           "description": "Compilation flags."
         },
+        "objectscript.searchAllDocTypes": {
+          "type": "boolean",
+          "default": false,
+          "description": "Whether server side search should search across all Studio Document types. Default is to only search classes, routines and include files."
+        },
         "objectscript.overwriteServerChanges": {
           "type": "boolean",
           "default": false,
--- a/package.json
+++ b/package.json
@@ -836,7 +836,7 @@
         "objectscript.searchAllDocTypes": {
           "type": "boolean",
           "default": false,
-          "description": "Whether server side search should search across all Studio Document types. Default is to only search classes, routines and include files."
+          "markdownDescription": "Whether [Quick Open](https://code.visualstudio.com/docs/getstarted/tips-and-tricks#_quick-open) should search across all Studio Document types. Default is to only search classes, routines and include files."
         },
         "objectscript.overwriteServerChanges": {
           "type": "boolean",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -833,6 +833,11 @@
           "default": "cuk",
           "description": "Compilation flags."
         },
+        "objectscript.searchAllDocTypes": {
+          "type": "boolean",
+          "default": false,
+          "description": "Whether server side search should search across all Studio Document types. Default is to only search classes, routines and include files."
```
gjsjohnmurray, line 839: I suggest making this a `markdownDescription` so it can use the term "Quick Open" instead of "server side search", and can hyperlink "Quick Open" to https://code.visualstudio.com/docs/getstarted/tips-and-tricks#_quick-open

src/providers/WorkspaceSymbolProvider.ts
```diff
@@ -2,7 +2,7 @@ import * as vscode from "vscode";
 import { AtelierAPI } from "../api";
 import { ClassDefinition } from "../utils/classDefinition";
 import { DocumentContentProvider } from "./DocumentContentProvider";
-import { StudioOpenDialog } from "../queries";
```
gjsjohnmurray, line 5: Why did you decide to remove this and also the type declaration of the result from the StudioOpenDialog query later?

gjsjohnmurray (changes requested):

package.json
isc-kiyer: Ah thanks! I did not know what the correct term was. Will make the update

src/providers/WorkspaceSymbolProvider.ts
isc-kiyer: @gjsjohnmurray The prior version of the query used "CALL". I switched it to "SELECT TOP 10 Name" because we only use the "Name" field. Should I revert to using CALL?

gjsjohnmurray: Sorry, I hadn't spotted that difference. No, I don't think there's any need to revert to using CALL. Reducing to a single column the amount of data returned from the query makes good sense, as does the use of SELECT TOP 10 (though that probably results in a change of behaviour in some situations, i.e. fewer results). @daimor may have an opinion about whether to still declare the result as StudioOpenDialog (not quite accurate), or to export another interface from query.d.ts that only contains Name.

isc-kiyer: @gjsjohnmurray @daimor Are there any further action items I should handle or are my changes sufficient?

daimor (approved):

gjsjohnmurray (approved):

PR #550, opened 2021-01-11
downgraded engine version requirements for katacoda

daimor:
This PR fixes #


gjsjohnmurray (approved):
Let's just hope we're not using anything that got introduced between 1.43 and 1.46

PR #560, opened 2021-01-27
Symbol search should include system files (% items)

isc-kiyer:
This PR fixes a backwards incompatible change made in #534. Prior to the change, Cmd/Ctrl + T would show system classes but after the change, those classes were no longer shown.


gjsjohnmurray:
@isc-kiyer a close look at the code prior to #534 makes me think that it used to show %-classes but not %-routines. I haven't gone as far as to verify that by testing with an older release. But if that's true, won't the change you propose here mean that %-routines get listed as well as %-classes?

isc-kiyer:
> @isc-kiyer a close look at the code prior to #534 makes me think that it used to show %-classes but not %-routines. I haven't gone as far as to verify that by testing with an older release. But if that's true, won't the change you propose here mean that %-routines get listed as well as %-classes?

@gjsjohnmurray Ah yes that is right. Would that be a problem? I think it is quite beneficial to open % include files in particular (both ones contained in IRISLIB and customer written ones beginning with %Z). Thoughts?

gjsjohnmurray:
@isc-kiyer I guess we should merge this and see. If anyone complains about the change we could add a setting.

@daimor OK with you? If so, please approve this PR.

daimor (approved):

PR #596, opened 2021-03-30
mapped local,global to private,public

### LINKED ISSUE #482: Change headings in Variables section of the Debugger view
In the Variables section of the debugging view, the current variables are shown in groupings that use the headings “local” and “global.” Since obviously “global” means something special to us, it would be great if the headings could be changed to “private” and “public”.

isc-rsingh:
This PR fixes #482

### DIFF
--- a/src/debug/debugSession.ts
+++ b/src/debug/debugSession.ts
@@ -82,6 +82,8 @@ export class ObjectScriptDebugSession extends LoggingDebugSession {
 
   private _contexts = new Map<number, xdebug.Context>();
 
+  private _contextNames: string[] = ["Private", "Public"];
+
   private _properties = new Map<number, xdebug.Property>();
 
   private _evalResultProperties = new Map<number, xdebug.EvalResultProperty>();
@@ -387,7 +389,7 @@ export class ObjectScriptDebugSession extends LoggingDebugSession {
     scopes = contexts.map((context) => {
       const variableId = this._variableIdCounter++;
       this._contexts.set(variableId, context);
-      return new Scope(context.name, variableId);
+      return new Scope(this._contextNames[context.id], variableId);
     });
     response.body = {
       scopes,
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (approved):
LGTM

daimor (approved):

gjsjohnmurray (changes requested):
IIRC (I am away from the codebase right now) the server side is theoretically able to send results with id=2. And maybe in future a 3 will get added. So I suggest you code defensively and fall back to using the name property if id is not less than the length of your string array.

isc-bsaviano:
@gjsjohnmurray @isc-rsingh The server can send context = 2 (Class).

gjsjohnmurray (approved):
Thanks @isc-rsingh.

@isc-bsaviano I spotted this comment in the context_names method of %Atelier.v2.XDebugAgent when trying to work out how to get the id=2 case to occur:
```objectscript
        
        #; Class is dependent upon existence of $this?.N1"@".E
        Set tResult="",tRet=..Debug.Eval("$this",.tResult)
        
        If tResult?.N1"@".E {
            Set tRes=tRes_"<context name='Class' id='"_..#CLASS_"'/>"
        }
```
Nowadays (2020.1) the `$this` SVN returns only the class name ('write $this` while in an instance method displays `YJM.Class1`). When I repeat this test on a 2010.2 environment I've got $this returns, for example, `1@YJM.Class1`. So it looks like something changed after 2010.2 to break this.

isc-bsaviano:
@gjsjohnmurray Jamie and I just tested the behavior of $this on an internal 2021 build and it works as intended (OREF in instance method). We even saw the Class context display the correct information in the debug view.

gjsjohnmurray:
@isc-bsaviano you are correct. My testing methodology was faulty. Sorry for the trouble.

isc-bsaviano:
@gjsjohnmurray No problem, glad it's not a bug.

PR #598, opened 2021-04-01
Capture children of eval command property response if they exist

isc-bsaviano:
We're considering adding support for drilling down into OREF's during debugging in the Atelier debugger so we'll need this code back into effect to get the most of the server enhancements.


daimor (approved):

gjsjohnmurray:
@isc-bsaviano is there any reason this shouldn't be merged? If yes then I suggest you convert it to draft. But if it's ready to merge and become the next beta then either @daimor or I can do this.

isc-bsaviano:
@gjsjohnmurray This is ready for merge. I would have merged it myself once Dmitry approved it but I don't have permissions for that.

PR #609, opened 2021-04-16
Add Document Links in ObjectScript Output Window

### LINKED ISSUE #386: Add a "Go to Error" functionality that parses error messages
Perhaps would be better as a separate issue, but are there any plans for a "Go to Error" functionality that parses error messages like @mvhenderson noted?

_Originally posted by @ajpino in https://github.com/intersystems-community/vscode-objectscript/issue_comments/697595493_

ty-d:
This PR fixes #386

Added links to classes and routines when errors occur in the ObjectScript output window. The links open to the corresponding file on isfs.

For example, clicking on ExamplePackage.Class.cls(ExampleMethod+5) will open the class and move the cursor to 5 lines below ExampleMethod.


gjsjohnmurray (changes requested):
The DocumentSymbolProvider you are invoking directly is nowadays only used if the InterSystems Language Server isn't present. I suggest you look at using the `vscode.executeDocumentSymbolProvider` API command instead, as documented [here](https://code.visualstudio.com/api/references/commands#commands). But be aware that the "(returns)" descriptions for the first two entries in that section are currently transposed (see https://github.com/microsoft/vscode/issues/121509).

Implementing your DocumentLinkProvider this way will avoid you having to replicate the "who to use" logic (LS or built-in), at the expense of some performance overhead. Providing links isn't going to be happening very often, so probably not a concern.

gjsjohnmurray (approved):
Looks good. Thanks for making the changes.

PR #685, opened 2021-06-14
fix #683 better isfs handling of server-side changes

### LINKED ISSUE #683: Unexpected alerts about server-side copy being newer when working with isfs
1. Have a workspace with an isfs root folder
2. Create a new routine (e.g. YJM2.int)
3. Save it.
4. Close it.
5. Without using VS Code, change the routine on the server.
6. Open the routine in the same workspace.
7. Save it (no need to change it in VS Code)
:bug: 
image

gjsjohnmurray:
This PR fixes #683

- Updates workspace-cached mtime when file is opened again
- Revises stat() method handling to detect when server-side etag has changed, triggering refresh of client cache entry.
- Enhances the 'failed compilation' error messages with a button to show the output channel.

@isc-bsaviano this PR should mean the work you proposed doing to detect server-side changes is no longer needed. Please try the VSIX this PR builds, or point relevant internal users at it for verification.

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -338,6 +338,11 @@ export class AtelierAPI {
         return this.cookies;
       }
 
+      // Not Modified
+      if (response.status === 304) {
+        throw { statusCode: response.status, message: response.statusText };
+      }
+
       const buffer = await response.buffer();
       const data: Atelier.Response = JSON.parse(buffer.toString("utf-8"));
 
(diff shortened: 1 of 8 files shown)


isc-bsaviano:
@gjsjohnmurray This looks good, but there's one weird behavior I noticed that may be unrelated to this change. When trying to save an ISFS file that has been updated on the server (the current dirty version is out of date), I get prompted with a message that gives me 2 options, "Compare" and "Overwrite". Once I clicked "Overwrite", a second message appeared with a similar text but this time had 4 options (this is the prompt that I get when editing client-side). Is it possible to remove that first message?

isc-bsaviano:
@gjsjohnmurray Just tested your update and the workflow is much better now. Clicking "Overwrite" performs the overwrite instead of going to the second popup message

isc-bsaviano (approved):

PR #686, opened 2021-06-14
fix #682 always run isfs dialog's serverInfo request in %SYS

### LINKED ISSUE #682: server does not have specified namespace
When [Configuring for Server-side Editing](https://intersystems-community.github.io/vscode-objectscript/serverside/#config-server-side) option for choosing namespace is not presented after selecting server.

Instead this error occurs "This server does not have specified namespace 'ETXX'. You must select one of the following: %SYS, ENSEMBLE, USER, WRC933601, WRC945238."

ETXX is being picked up from objectscript.conn, but why should that be? (server is different).

 "objectscript.conn": {
        "server": "healthconnect20191vscuser",
        "ns": "ETXX",
        "active": true
    }

gjsjohnmurray:
This PR fixes #682

### DIFF
--- a/src/commands/addServerNamespaceToWorkspace.ts
+++ b/src/commands/addServerNamespaceToWorkspace.ts
@@ -24,7 +24,7 @@ export async function addServerNamespaceToWorkspace(): Promise<void> {
     return;
   }
   // Get its namespace list
-  let uri = vscode.Uri.parse(`isfs://${serverName}/`);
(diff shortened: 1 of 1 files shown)


gjsjohnmurray:
Prior to this change it would use the objectscript.conn.ns value even if objectscript.conn.server pointed to a different server.

isc-bsaviano (approved):

PR #715, opened 2021-07-09
Allow configuration of server side search max results

### LINKED ISSUE #713: Server side search has a default max of ~200 results. This should be configurable
When using server side search using the proposed APIs, search results are limited to just over 200. This setting should be configurable so more results can be returned if desired:

Extension version: v1.0.13-beta3
VS code version: 1.57.1

Per @isc-bsaviano this is due to the REST request using a default value of 200. A setting should be exposed to override this default.

isc-bsaviano:
This PR fixes #713 

Edited `provideTextSearchResults()` to send `options.maxResults` to the server if it's non-`null`. If the value is `null` (meaning "send an unlimited number of results"), a value of 100,000 is sent to the server because it's not possible to specify "unlimited".

### DIFF
--- a/src/providers/FileSystemProvider/TextSearchProvider.ts
+++ b/src/providers/FileSystemProvider/TextSearchProvider.ts
@@ -29,6 +29,9 @@ export class TextSearchProvider implements vscode.TextSearchProvider {
         regex: query.isRegExp,
         word: query.isWordMatch,
         case: query.isCaseSensitive,
+        // If options.maxResults is null the search is supposed to return an unlimited number of results
+        // Since there's no way for us to pass "unlimited" to the server, I choose a very large number
+        max: options.maxResults !== null ? options.maxResults : 100000,
       })
       .then((data) => data.result)
       .then((files: SearchResult[]) =>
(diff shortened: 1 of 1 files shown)


src/providers/FileSystemProvider/TextSearchProvider.ts
```diff
@@ -29,6 +29,9 @@ export class TextSearchProvider implements vscode.TextSearchProvider {
         regex: query.isRegExp,
         word: query.isWordMatch,
         case: query.isCaseSensitive,
+        // If options.maxResults is null the search is supposed to return an unlimited number of results
+        // Since there's no way for us to pass "unlimited" to the server, I choose a very large number
+        max: options.maxResults !== null ? options.maxResults : 100000,
```
gjsjohnmurray, line 34:
```suggestion
        max: options.maxResults ?? 100000,
```

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing

gjsjohnmurray (changes requested):

src/providers/FileSystemProvider/TextSearchProvider.ts
isc-bsaviano: Cool, I didn't know that

gjsjohnmurray (approved):

PR #716, opened 2021-07-09
Require confirmation for namespace compile command

### LINKED ISSUE #677: Compile all namespace action needs confirmation
It's not so good having this action with no confirmation at all, it should ask if a user really wants to compile entire namespace.

isc-bsaviano:
This PR fixes #677

### DIFF
--- a/src/commands/compile.ts
+++ b/src/commands/compile.ts
@@ -337,6 +337,15 @@ export async function namespaceCompile(askFlags = false): Promise<any> {
   if (!config("conn").active) {
     throw new Error(`No Active Connection`);
   }
+  const confirm = await vscode.window.showWarningMessage(
+    `Compiling all files in namespace '${api.ns}' is expensive! Are you sure you want to proceed?`,
+    "Cancel",
+    "Confirm"
+  );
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (approved):

src/commands/compile.ts
```diff
@@ -337,6 +337,15 @@ export async function namespaceCompile(askFlags = false): Promise<any> {
   if (!config("conn").active) {
     throw new Error(`No Active Connection`);
   }
+  const confirm = await vscode.window.showWarningMessage(
+    `Compiling all files in namespace '${api.ns}' is expensive! Are you sure you want to proceed?`,
```
gjsjohnmurray, line 341:
```suggestion
    `Compiling all files in namespace '${api.ns}' might be expensive. Are you sure you want to proceed?`,
```

gjsjohnmurray (changes requested):
Second thoughts about the dialog wording.

src/commands/compile.ts
isc-bsaviano: I like your change but I merged the PR before you made it. Can you commit it directly?

PR #730, opened 2021-08-16
Add GUI editors for BPL and DTL files

### LINKED ISSUE #60: Add the support for BPL and DTL files
BPL stands for Business Process, DTL - for Data transformation rules.
Both are classes which have GUI editors in IRIS.

ty-d:
This PR fixes #60

Notes:
- **This feature is not ready to be merged into the extension.**
- The BPL/DTL editors will not work without changes to server side classes, which are being worked on outside of this PR
- This requires VS Code version 1.59 because of webview bugs that caused SVGs to not load.

This adds a custom editor for .bpl and .dtl files. Whenever those files are opened, the GUI Zen editor will be opened allowing the user to edit and save the class, which automatically updates the class version. Edits to the class version will cause the Zen version to reload with the new changes when the user clicks on it, unless the Zen version has also been edited in which case it will ask the user whether to reload or not. The view other command can be used to switch between the .cls and .bpl/.dtl versions.

A class that has been exported to the local filesystem will be updated by the Zen editor in the same way, so client side source control should not be an issue.

Keyboard shortcuts implemented in the webview:
- Ctrl + S = save the document
- Ctrl + Z = undo
- Ctrl + Shift + V = view other (open the class version)

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     }
   ],
   "engines": {
-    "vscode": "^1.43.0"
+    "vscode": "^1.59.0"
   },
   "enableProposedApi": true,
   "activationEvents": [
--- a/package.json
+++ b/package.json
@@ -1059,6 +1059,21 @@
         ]
       }
     ],
+    "customEditors": [
+      {
+        "viewType": "vscode-objectscript.bplDtlEditor",
+        "displayName": "BPL Editor",
+        "selector": [
+          {
+            "filenamePattern": "*.bpl"
+          },
+          {
+            "filenamePattern": "*.dtl"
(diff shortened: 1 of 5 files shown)


isc-bsaviano (changes requested):
@ty-d Please update this line in `package.json` to reflect the new minimum version of `1.59` according to the PR description

https://github.com/intersystems-community/vscode-objectscript/blob/ff371902194ad24ac8b3d568be29a9a8154ce532/package.json#L49-L51

gjsjohnmurray:
PMC decided unanimously to close this without merging.

PR #739, opened 2021-09-23
security fix for path-parse

isc-rsingh:
This PR fixes dependabot alerts

### DIFF
--- a/package.json
+++ b/package.json
@@ -1136,6 +1136,7 @@
     "webpack-cli": "^4.5.0"
   },
   "dependencies": {
+    "@xmldom/xmldom": "^0.7.5",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1148,7 +1149,6 @@
     "vscode-debugadapter": "^1.41.0",
     "vscode-debugprotocol": "^1.41.0",
     "vscode-extension-telemetry": "^0.1.6",
-    "ws": "^7.4.6",
```
daimor, line 1151: Why did you remove 'we'?

isc-rsingh: It's still there. Down lower.

PR #743, opened 2021-09-30
Fix XML to UDL conversion

isc-bsaviano:


gjsjohnmurray (approved):
LGTM but I haven't tested it.

PR #760, opened 2021-11-02
Refresh ObjectScript Explorer after export

### LINKED ISSUE #502: Visual file indicators/colors do not update after ObjectScript Explorer -  option Export
Visual file indicators/colors do not update after ObjectScript Explorer - option Export (See Snapshots below)

Here is a file I just exported (note the blue decoration on the left and the U at the end of the column is missing):

 
vsc-171l

If I click on the file above it (which I think forces a refresh), they now appear:

 
vsc-171m

So it appears there is no refresh happening after the Export.  InterSystems Support also confirms this behavior.

isc-bsaviano:
This PR fixes #502

### DIFF
--- a/src/commands/export.ts
+++ b/src/commands/export.ts
@@ -2,7 +2,7 @@ import fs = require("fs");
 import path = require("path");
 import * as vscode from "vscode";
 import { AtelierAPI } from "../api";
-import { config } from "../extension";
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (approved):
LGTM

PR #761, opened 2021-11-02
Add auto-commenting of new lines following existing ObjectScript comments

### LINKED ISSUE #541: Better commenting for Cache'
The one feature I miss from Cache' Studio is the automagic indenting and commenting.  It's very nice to not have to retype the comment character when I'm writing a long series of full-line comments (e.g. a tag header).  When setting up a new Cache' Studio instance, I always make a point to set the User Defined indentation options in
          Tools -> Options -> Editor -> Indentation 

I tried using the "Auto Comment Next Line" extension (https://github.com/ctf0/vscode-auto-comment-next-line) to duplicate this Cache' Studio functionality, but it didn't work correctly.  The problem was that the code in vscode-objectscript/src/languageConfiguration.ts  contains a lineComment setting that doesn't support raw MUMPS.  Instead of a ";" character, the lineComment setting only contains "#;" token.  This means that when I try to use something like the , I keep getting "#;" tokens blown in as my leading characters.

It might be nice to put in some kind of OnEnterRules that can carry over the leading characters of the previous line, based on a user-defined list.  This would exactly duplicate the Cache' Studio functionality.

isc-bsaviano:
This PR fixes #541

### DIFF
--- a/src/languageConfiguration.ts
+++ b/src/languageConfiguration.ts
@@ -19,10 +19,34 @@ export function getLanguageConfiguration(lang: string): LanguageConfiguration {
         ? [
             {
               beforeText: /^\/\/\//,
-              afterText: /.*/,
               action: { indentAction: IndentAction.None, appendText: "/// " },
             },
           ]
-        : [],
+        : [
+            {
+              beforeText: /^\s\/\/\//,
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (approved):
Let's merge this and see if anyone complains it's made something worse.

PR #770, opened 2021-11-12
Fix `FileSystemProvider` mtime caching

isc-bsaviano:
This PR fixes `FileSystemProvider` mtime caching. Previously, the cache would get constantly overwritten so calls to `GET /doc` wouldn't include the `IF_NONE_MATCH` header, forcing the server to send the full contents of the file back every time. This fix corrects this problem so the proper mtime is sent with `GET /doc` requests.

### DIFF
--- a/src/providers/FileSystemProvider/FileSystemProvider.ts
+++ b/src/providers/FileSystemProvider/FileSystemProvider.ts
@@ -88,7 +88,9 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
             const name = item.Name;
             const fullName = folder === "" ? name : csp ? folder + name : folder + "/" + name;
             if (item.Type === "10" || item.Type === "9") {
-              parent.entries.set(name, new Directory(name, fullName));
+              if (!parent.entries.has(name)) {
+                parent.entries.set(name, new Directory(name, fullName));
+              }
               return [name, vscode.FileType.Directory];
             } else {
               return [name, vscode.FileType.File];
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (approved):
Based on code-reading this looks OK to me. I haven't tested the dev VSIX though.

isc-bsaviano:
@gjsjohnmurray @daimor There's something wrong with the CI. Compile fails because it can't find the `vscode` module. I'm not going to merge this until we get a CI run for this PR that succeeds

daimor:
There is something wrong with downloading `vscode.proposed.d.ts`

daimor:
https://github.com/microsoft/vscode/issues/136964

gjsjohnmurray:
> 
> 
> [microsoft/vscode#136964](https://github.com/microsoft/vscode/issues/136964)

Correct. The CI needs to be revised to handle the relocation of vscode.d.ts and the breaking up of vscode.proposed.d.ts.

But I think the reason this is suddenly biting us is that the CI downloads these from main (actually, still using the old name 'master'). It should probably be using npx instead.

daimor:
I'll take care of this change, and create a PR for it

isc-bsaviano:
@gjsjohnmurray We have a dev vsix for this now if you want to test it before I merge

gjsjohnmurray:
A quick test of the dev vsix didn't find anything broken, so I'm happy for my approval to stand.

PR #771, opened 2021-11-12
fix for deprecated vscode.proposed.d.ts

daimor:
Build will fail until the vscode team will publish the latest vscode-dts


isc-bsaviano:
According to https://github.com/microsoft/vscode/issues/136964#issue-1051051480, this line:

https://github.com/intersystems-community/vscode-objectscript/blob/9dc677ad62190f064195586291cc33a9de80bf04/package.json#L52

should be changed to:

```json
  "enabledApiProposals": ["fileSearchProvider", "textSearchProvider"],
```

gjsjohnmurray (changes requested):
I think we should leave the old line in there as well, until after 1.63 ships as Stable in December. And even then only remove it if we bump our minimum required version to 1.63 (currently it is 1.43). Otherwise people trying to used the proposed APIs in current Stable (1.62.2) or earlier are likely to find they no longer work.

isc-bsaviano:
With https://github.com/microsoft/vscode-dts/pull/11 merged we will be able to continue to use `vscode-dts dev` without specifying the specific proposals

jrieken:
> With microsoft/vscode-dts#11 merged we will be able to continue to use vscode-dts dev without specifying the specific proposals

fyi - this has been merged and published and you can simply run `vscode-dts dev` BUT you do need to spell out each proposal in your extensions package.json file first, like so https://github.com/microsoft/vscode-anycode/blob/e04704aaf6d813c4e882f54dd65f04ff16b6c08b/package.json#L8

isc-bsaviano (approved):

gjsjohnmurray (approved):

PR #772, opened 2021-11-16
Overhaul `WorkspaceSymbolProvider`

isc-bsaviano:
- Search for all class members, not just methods.
- Don't search for Studio documents so we avoid duplication functionality that's covered by the `FileSearchProvider` (Quick Open).
- Properly translate `query` to a `%MATCHES` compatible pattern while following the interpretation guidelines.
- Use the same `SymbolKind` as the Language Server for all class member types for consistency.
- Use the `vscode.executeDocumentSymbolProvider` command to find the location of the symbol.
- Improve performance by only computing symbol location when selected by the user.
- Remove unneeded setting `"objectscript.searchAllDocTypes"`.

### DIFF
--- a/src/providers/WorkspaceSymbolProvider.ts
+++ b/src/providers/WorkspaceSymbolProvider.ts
@@ -7,23 +7,23 @@ export class WorkspaceSymbolProvider implements vscode.WorkspaceSymbolProvider {
   private sql: string =
     "SELECT * FROM (" +
     "SELECT Name, Parent->ID AS Parent, 'method' AS Type FROM %Dictionary.MethodDefinition" +
-    " UNION ALL " +
+    " UNION ALL %PARALLEL " +
     "SELECT Name, Parent->ID AS Parent, 'property' AS Type FROM %Dictionary.PropertyDefinition" +
-    " UNION ALL " +
+    " UNION ALL %PARALLEL " +
     "SELECT Name, Parent->ID AS Parent, 'parameter' AS Type FROM %Dictionary.ParameterDefinition" +
-    " UNION ALL " +
+    " UNION ALL %PARALLEL " +
     "SELECT Name, Parent->ID AS Parent, 'index' AS Type FROM %Dictionary.IndexDefinition" +
-    " UNION ALL " +
+    " UNION ALL %PARALLEL " +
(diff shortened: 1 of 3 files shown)


src/providers/WorkspaceSymbolProvider.ts
```diff
@@ -1,85 +1,91 @@
...
+        let kind: vscode.SymbolKind = vscode.SymbolKind.Property;
+        if (element.Type === "method" || element.Type === "query") {
+          kind = vscode.SymbolKind.Method;
+        } else if (element.Type === "parameter") {
+          kind = vscode.SymbolKind.Constant;
+        } else if (element.Type === "index") {
+          kind = vscode.SymbolKind.Key;
+        } else if (element.Type === "xdata" || element.Type === "storage") {
+          kind = vscode.SymbolKind.Struct;
+        }
```
daimor, line 54:
```suggestion
    const kind: vscode.SymbolKind = (() => {
      switch (element.type) {
        case "query":
        case "method":
          return vscode.SymbolKind.Method;
        case "parameter":
          return vscode.SymbolKind.Constant;
        case "index":
          return vscode.SymbolKind.Key;
        case "xdata":
        case "storage":
          return vscode.SymbolKind.Struct;
        case "property":
        default:
          return vscode.SymbolKind.Property;
      }
    })();
```
I think this will make it just more readable

src/providers/WorkspaceSymbolProvider.ts
```diff
@@ -1,85 +1,91 @@
...
 export class WorkspaceSymbolProvider implements vscode.WorkspaceSymbolProvider {
-  public provideWorkspaceSymbols(
-    query: string,
-    token: vscode.CancellationToken
-  ): vscode.ProviderResult<vscode.SymbolInformation[]> {
-    if (query.length < 3) {
+  private sql: string =
+    "SELECT * FROM (" +
+    "SELECT Name, Parent->ID AS Parent, 'method' AS Type FROM %Dictionary.MethodDefinition" +
+    " UNION ALL " +
```
daimor, line 10:
```suggestion
    " UNION ALL  %PARALLEL" +
```
And for all the next `UNION ALL`'s, it'll the whole query much faster.

daimor (changes requested):

daimor (approved):

PR #776, opened 2021-11-18
Append `.pkg` to package name when passed to source control / server command extensions

### LINKED ISSUE #774: Package names passed to source control / server command extensions are missing `.pkg` extension
Unlike Studio, names of packages selected by users in the explorer view are not passed with the `.pkg` extension in VS Code. This creates inconsistencies and failures in VS Code. Our current workaround for this is to check if the package has any classes and explicitly add the extension if it does. This will not work if the package has no classes. 

Also see: [this git-source-control hooks issue](https://github.com/intersystems/git-source-control/issues/71) 

@timleavitt

isc-bsaviano:
This PR fixes #774

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -238,7 +238,13 @@ export function getServerName(uri: vscode.Uri): string {
     return uri.path;
   } else {
     // Complex case: replace folder slashes with dots.
-    return uri.path.slice(1).replace(/\//g, ".");
+    const filePath = uri.path.slice(1);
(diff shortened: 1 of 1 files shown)


isc-bsaviano:
@isc-tleavitt @isc-svelury Please download the vsix for this PR [here](https://github.com/intersystems-community/vscode-objectscript/suites/4396058115/artifacts/115933577) and verify that it behaves as expected

isc-svelury:
Works as expected for me. Thank you!

daimor (approved):

PR #778, opened 2021-11-23
Add option to open shell in docker

grongierisc:
With the up coming version of IRIS with embedded python, it can be useful to have an easy option to gain shell access in the container


daimor (approved):

gjsjohnmurray (changes requested):
Maybe resolve the linting failure in utils.ts

daimor (approved):

PR #782, opened 2021-11-25
Added support for work in browser

daimor:
For cases like github.dev and vscode.dev, following the [recommendations](https://code.visualstudio.com/api/extension-guides/web-extensions) 
Very basic support, only coloring, and symbol providers

image


daimor:
@gjsjohnmurray what do you think about this PR?

package.json
```diff
@@ -49,7 +49,7 @@
   "engines": {
     "vscode": "^1.43.0"
   },
-  "enableProposedApi": true,
+  "enableProposedApi": false,
```
gjsjohnmurray, line 52: Why this change? I thought we set it true so beta builds work with proposed API, then the release action changes it to false. I know this property is going away at 1.63, but at the moment we declare 1.43 as our minimum version. And 1.63 hasn't shipped yet.

gjsjohnmurray (changes requested):

package.json
daimor: yeah, returned it back

gjsjohnmurray (approved):

PR #790, opened 2021-11-30
Prepare for 1.2.0 release

isc-bsaviano:


daimor (dismissed):

CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
 # Change Log
 
+## [1.1.2] 30-Nov-2021
+- Enhancements
+  - Overhaul `WorkspaceSymbolProvider` (#772)
+  - Add `Open Shell in Docker` option to Server Actions menu (#778)
+  - Premilinary web extension support (#782)
```
gjsjohnmurray, line 7:
```suggestion
  - Preliminary web extension support (#782)
```

gjsjohnmurray (changes requested):

gjsjohnmurray (approved):
LGTM, but maybe we need to wait until the issue reported today about not working with Insiders gets resolved. 1.63 started endgame yesterday and is likely to ship a week on Thursday.

isc-bsaviano:
@gjsjohnmurray Yes, I think we need to wait until that issue is resolved

gjsjohnmurray:
I propose we renumber this to 1.2. One reason is that #782 (github.dev and vscode.dev support) is a significant step. Another is I'd like us to be ready to use the upcoming extension pre-release mechanism (see https://github.com/microsoft/vscode/issues/138068). It's already working in 1.63 Insiders, which is now in endgame and expected to release next Thursday.

The recommended version numbering pattern for the pre-release mechanism is to use EVEN numbers in the minor position for the regular releases and ODD for pre-releases, so it'd be good to get the extension onto an even minor (2) soon.

Also pinging @isc-rsingh for an opinion on this.

daimor:
I like the idea with pre-releases, so, it probably will be possible to produce our beta releases this way directly to marketplace. Beta-tests users will get new versions very quickly.

isc-bsaviano:
@gjsjohnmurray Would it be possible to upload our beta versions in the pre-release stream using the current "-beta.X" suffix? If so, I think we should just modify that action to have a publish step and keep everything else as-is.

daimor:
Marketplace at the moment allows only three parts, major.minor.patch
So, beta version number will go to patch part, and normal releases will get 0 there

isidorn:
I see you are already looking into this, you can check out the docs which are work in progress and will be ready tomorrow
https://github.com/microsoft/vscode-docs/blob/vnext/api/working-with-extensions/publishing-extension.md#pre-release-extensions

gjsjohnmurray:
> beta version number will go to patch part

An alternative strategy would be to use YYYYMMDDNN as the patch digit for pre-releases. This format means it's easier to differentiate pre-releases from regular ones, and to see at a glance how recent a particular pre-release is. The NN part gives scope for up to 99 pre-releases in a single day, which should be more than enough!

isc-bsaviano:
@gjsjohnmurray @daimor Can you add the CI changes to this PR or should we open another one for that? Also, can pre-release versions published to the marketplace use the proposed API? If so, shouldn't we update our documentation?

isc-bsaviano:
@daimor I will wait for the CI updates to be decided upon before I update and merge this PR because I will likely have to make README changes

isc-rsingh:
I also agree with calling this 1.2 and all your reasoning behind it @gjsjohnmurray

PR #800, opened 2021-12-06
fix default values for export

### LINKED ISSUE #799: Exporting not working with new version 1.2.1
Hi,

I exported 2 new tables (classes) from the Object Script Explorer after creating them with DDL syntax and they were not appearing as usual. This is my configuration:
"objectscript.export": {
"folder": "",
"addCategory": false,
},

I realized then than instead of exporting the class inside the folder tree of its package (namespace), it was exporting it to the root using the whole package in its name. I rolledback the extension to the previous version, 1.1.1, and it worked like always during last year. I use this functionality nearly every week. Thanks for the extension.

Best,
Jose

daimor:
This PR fixes #799

### DIFF
--- a/package.json
+++ b/package.json
@@ -831,6 +831,12 @@
           "type": "object",
           "description": "Control what to export from the server into the local folder.",
           "scope": "resource",
+          "default": {
+            "folder": "src",
+            "addCategory": false,
+            "map": {},
+            "atelier": true
+          },
           "properties": {
             "folder": {
               "description": "Folder for exported source code within workspace.",


isc-bsaviano:
@daimor Why didn't you include defaults for all of the export settings?

daimor:
Left, just most of the important and which are not just false by default.

isc-bsaviano:
Yes but those will be `undefined` now if they aren't overridden, just like the ones that caused #799. I don't see why you changed the package.json for these settings in #782 and now that we know we need to specify the defaults this way I don't see a good reason to not explicitly set all of the defaults. We should probably delete the old defaults as well since they don't work anymore.

daimor:
Yeah, I changed it, because I saw some errors, in VSCode logs, something like settings overridden, so, had to put them in correct order. Yeah, probably we would need to delete those defaults, as the useless now. But I see no reasons to put, rarely used parameters in default. With false values, while they supposed to be used as booleans and undefined works as false.

PR #801, opened 2021-12-06
Restore defaults for all export settings

### LINKED ISSUE #799: Exporting not working with new version 1.2.1

isc-bsaviano:
This PR improves #800 and fixes #799

@daimor @gjsjohnmurray I think we should do a 1.2.2 release once this is merged because #799 is causing users to downgrade

### DIFF
--- a/package.json
+++ b/package.json
@@ -835,26 +835,29 @@
             "folder": "src",
             "addCategory": false,
             "map": {},
-            "atelier": true
+            "atelier": true,
+            "generated": false,
+            "filter": "",
+            "category": "*",
+            "noStorage": false,
+            "dontExportIfNoChanges": false,
+            "maxConcurrentConnections": 0
           },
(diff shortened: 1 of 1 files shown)


daimor (approved):

gjsjohnmurray (approved):
Looks fine. I haven't tested it, but between the two of you I assume you have covered everything this time.

Issue #781, opened 2021-11-24
Attempted Edit


AndreaSchecker:
Issue Type: <b>Bug</b>

I programmed some server command by my own. That server commands worked fine for a long time.
Suddonly I received 
Studio Action "Attempted Edit" not supported on 127.0.0.1:57772[CTAE01]
by using one of my server commands.
Explanation:
Error running command vscode-objectscript.serverCommands.contextOther: Running the contributed command: 'vscode-objectscript.serverCommands.contextOther' failed.. This is likely caused by the extension that contributes vscode-objectscript.serverCommands.contextOther.

Extension version: 1.1.1
VS Code version: Code 1.61.2 (6cba118ac49a1b88332f312a8f67186f7f3c1643, 2021-10-19T14:57:20.575Z)
OS version: Windows_NT x64 10.0.19042
Restricted Mode: No

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz (4 x 2904)|
|GPU Status|2d_canvas: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>oop_rasterization: enabled<br>opengl: enabled_on<br>rasterization: enabled<br>skia_renderer: enabled_on<br>video_decode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled|
|Load (avg)|undefined|
|Memory (System)|31.90GB (19.14GB free)|
|Process Argv|. --crash-reporter-id 9812a2f4-d7e0-49de-ac19-9cf9beb0ed10|
|Screen Reader|no|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368:30146709
vsreu685:30147344
python383cf:30185419
vspor879:30202332
vspor708:30202333
vspor363:30204092
vswsl492cf:30256860
pythontb:30283811
pythonptprofiler:30281270
vshan820:30294714
vstes263:30335439
pythondataviewer:30285071
vscod805cf:30301675
pythonvspyt200:30340761
binariesv615:30325510
bridge0708:30335490
dockerwalkthru:30377721
bridge0723:30353136
pythonrunftest32:30373476
pythonf5test824:30373475
javagetstartedt:30391933
pythonvspyt187:30373474
vsqsis200cf:30402084
vsaa593:30376534
vssld246cf:30401891
vscexrecpromptt3:30400063
[...]

PR #814, opened 2021-12-27
Fix proposed API registration

isc-bsaviano:
This PR fixes an issue where beta and dev vsix's cannot be activated with proposed APIs disabled.

See #781, starting from https://github.com/intersystems-community/vscode-objectscript/issues/781#issuecomment-997830201 for how this issue manifests.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -634,24 +634,20 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
 
   const diagnosticProvider = new ObjectScriptDiagnosticProvider();
 
-  // Gather the proposed APIs we will register to use when building with enableProposedApi = true
+  // Gather the proposed APIs we will register to use when building with enabledApiProposals != []
   const proposed = [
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (commented):
Are the `packageJson.enableProposedApi` tests still needed?

isc-bsaviano:
@gjsjohnmurray Nope, just tested without them and everything still works. I will remove them.

daimor (approved):

gjsjohnmurray:
With this change does the argv json setting really become redundant, as Gabriel's comment on the issue implies? I am away from my desk so unable to test.

isc-bsaviano:
@gjsjohnmurray There's more context to what he is saying. The issue we ran into is that activation failed when the argv setting wasn't present, when the expected behavior is that all non-proposed features would be enabled and the extension would activate successfully. We still expect that the argv setting is required to use the proposed APIs, but we expect that with this PR it won't be required to activate a beta or dev vsix.

PR #815, opened 2021-12-28
Provide `WorkspaceSymbols` for all servers in workspace

### LINKED ISSUE #375: Multi-root workspace: 'Open Symbol by Name' only searches one namespace
If a workspace connects to more than one namespace (same or different servers), the `Ctrl/Cmd+T` lookup only searches one. I think it should act on all namespaces the workspace links to.

Noticed when fixing #373.

isc-bsaviano:
This PR fixes #375

Changes made:
- Loop through folders in the workspace and determine which server and namespace they’re connected to.
- Filter out duplicate server/namespace combos to avoid querying a server multiple times for the same namespace.
- Filter out results from system classes in non-%SYS namespaces when a workspace is connected to both the %SYS and a non-%SYS namespace for the same server to avoid duplicate results.

### DIFF
--- a/src/providers/WorkspaceSymbolProvider.ts
+++ b/src/providers/WorkspaceSymbolProvider.ts
@@ -1,6 +1,5 @@
 import * as vscode from "vscode";
 import { AtelierAPI } from "../api";
-import { currentWorkspaceFolder } from "../utils";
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (approved):
Looks good but I haven't tested it.

PR #818, opened 2021-12-30
Improve exporting

### LINKED ISSUE #129: Need ability to export currently open artifact
From VSCode, we can use the ObjectScript Explorer to navigate on the server to the artifact, right-click on it, and select Export. This adds the artifact locally. This works fine.

However, we have a TON of routines. Like, tens of thousands. Using the explorer to find all the ones we want to add locally is difficult.

Feature suggestion: Ability to do so from the VSCode Command Palette. We should be able to either invoke, for example, >ObjectScript: Export ClassName to export a specific artifact that isn't open, or >ObjectScript: Export and it would export the currently open artifact that we've opened using Ctrl-T.

isc-bsaviano:
This PR fixes #129

Changes made:
- Modify `"vscode-objectscript.export"` to do the following:
    - Don’t allow the user to select a virtual folder as the export destination.
    - If there are no suitable folders in the workspace, notify the user and exit.
    - Replace the call to `AtelierAPI.getDocNames()` with a call to the StudioOpenDialog query to improve performance and to allow the filtering out of system items when the namespace is not `%SYS`.
    - Ask the user for confirmation and allow them to exclude individual files from the export list before starting the export.
- Add `"vscode-objectscript.exportCurrentFile”` command that exports the currently opened file, if it was opened from the ObjectScript Explorer.
- Change `RootNode.getList()` query to only SELECT needed columns from StudioOpenDialog to improve performance.
- Improve documentation for `"objectscript.export.filter"` setting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -506,7 +506,7 @@
       {
         "category": "ObjectScript",
         "command": "vscode-objectscript.export",
-        "title": "Export Files",
+        "title": "Export Code from Server",
         "enablement": "!isWeb"
       },
       {
@@ -708,7 +708,7 @@
       {
         "category": "ObjectScript",
         "command": "vscode-objectscript.exportCurrentFile",
-        "title": "Export Current File",
+        "title": "Export Current File from Server",
         "enablement": "!isWeb"
       }
     ],
--- a/package.json
+++ b/package.json
@@ -233,6 +233,10 @@
         {
           "command": "vscode-objectscript.showClassDocumentationPreview",
           "when": "editorLangId == objectscript-class"
+        },
+        {
+          "command": "vscode-objectscript.exportCurrentFile",
+          "when": "resourceScheme == objectscript && vscode-objectscript.connectActive"
         }
       ],
       "view/title": [
@@ -333,6 +337,11 @@
           "command": "vscode-objectscript.compileOnly",
           "when": "editorLangId =~ /^objectscript/ && vscode-objectscript.connectActive",
           "group": "objectscript@7"
+        },
+        {
+          "command": "vscode-objectscript.exportCurrentFile",
+          "when": "resourceScheme == objectscript && vscode-objectscript.connectActive",
+          "group": "objectscript@8"
         }
       ],
       "editor/title": [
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -695,6 +704,12 @@
         "title": "Show Class Documentation Preview",
         "enablement": "!isWeb",
         "icon": "$(open-preview)"
+      },
+      {
+        "category": "ObjectScript",
+        "command": "vscode-objectscript.exportCurrentFile",
+        "title": "Export Current File",
```
gjsjohnmurray, line 711: Related to my comment above, but I'm finding it hard to come up with a title I'm happy with. Maybe "Export Server Version of Current File".

src/commands/export.ts
```diff
@@ -216,27 +209,86 @@ export async function exportList(files: string[], workspaceFolder: string, names
...
+    });
+    if (selection === undefined) {
+      return;
     }
+    workspaceFolder = selection;
+  } else if (workspaceList.length === 1) {
+    workspaceFolder = workspaceList.pop();
+  } else {
+    vscode.window.showInformationMessage(
+      "There are no folders in the current workspace that files can be exported to.",
```
gjsjohnmurray, line 229: "files" -> "server code" or simply "code", if my other suggestion is adopted.

src/commands/export.ts
```diff
@@ -216,27 +209,86 @@ export async function exportList(files: string[], workspaceFolder: string, names
...
+      filterStr,
+    ])
+    .then(async (data) => {
+      let files: vscode.QuickPickItem[] = data.result.content.map((file) => {
+        return { label: file.Name, picked: true };
+      });
+      files = await vscode.window.showQuickPick(files, {
+        canPickMany: true,
+        ignoreFocusOut: true,
+        placeHolder: "Click a file to exclude it. Press 'Escape' to cancel export.",
```
gjsjohnmurray, line 280:
```suggestion
        placeHolder: "Uncheck a file to exclude it. Press 'Escape' to cancel export.",
```

gjsjohnmurray (changes requested):
Various comments for consideration.

gjsjohnmurray (approved):

PR #822, opened 2022-01-13
Add default file icons

isc-bsaviano:
https://github.com/microsoft/vscode/issues/140047 adds the ability to provide default file icons for contributed languages. This PR adds the InterSystems logo as the default file icon for all of our contributed languages (classes, routines, includes and CSP/CSR). Unfortunately, the Salesforce cloud will still be the file icon for `.cls` files because it comes from the default [Seti file icon theme](https://github.com/jesseweed/seti-ui). Once this PR gets approved we should consider opening a PR in that repo that uses this icon for `.mac`, `.int`,  `.inc`, `.csp` and `.csr` files so nobody else beats us to them.

EDIT: I've included some screenshots below.

This is the current behavior:
Screen Shot 2022-01-13 at 10 07 59 AM
This is with this PR using a dark theme:
Screen Shot 2022-01-13 at 1 20 54 PM
This is with this PR using a light theme:
Screen Shot 2022-01-13 at 1 21 08 PM

### DIFF
--- a/images/fileIcon.svg
+++ b/images/fileIcon.svg
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="32px" height="32px" viewBox="0 0 24 32" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 61.2 (89653) - https://sketch.com -->
+    <title>Product icon</title>
+    <desc>Created with Sketch.</desc>
+    <defs>
+        <polygon id="path-1" points="0 0.90625 7.25 0.90625 7.25 27.1875 0 27.1875"></polygon>
+    </defs>
+    <g id="Symbols" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="components/header-no-burger" transform="translate(-7.000000, -10.000000)">
+            <g id="left" transform="translate(-1.000000, 4.000000)">
+                <g id="logos/icon-HS" transform="translate(8.000000, 6.000000)">
+                    <g id="Icon" stroke-width="1">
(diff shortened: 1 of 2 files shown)


isc-rsingh (dismissed):

daimor (approved):

gjsjohnmurray:
Did you consider using the same icons as Studio, which would distinguish INC from MAC/IMT?

gjsjohnmurray:
Also, won't this require 1.64?

isc-bsaviano:
@gjsjohnmurray No, I didn't consider using those icons but I will look into it before merging this. Yes, this requires 1.64 to actually show the icons but when I tested on 1.63 everything worked the same as before so I think this is safe to merge before 1.64 is out.

isc-bsaviano:
@gjsjohnmurray I fired up Studio again to get a look at the icons and I don't think they're usable. I agree that it would be nice to have different icons for every file extension but since that's not possible in the `package.json` file I think this is good enough for now. If/when we decide to add icons to the Seti theme we can consider how to differentiate the different file types.

gjsjohnmurray:
@isc-bsaviano each language contributed by the extension can specify a different icon.

At the moment we treat INTs and MACs as the same language. IMO this is a mistake, because it means I can write in my INTs code which is only valid in MACs (e.g. #-directives, embedded SQL etc) and only find out when I try to save/import them into a namespace.

To fix this I think we should add a new language `objectscript-int` and associate the .int extension with this rather than with `objectscript`. Then in this PR INTs and MACs can be given different icons.

I will open a separate issue for this, which I suggest should get priority over this PR.

gjsjohnmurray:
@isc-bsaviano now that #865 has merged I suggest you merge master into your branch, update package.json to specify icons for the new `objectscript-int` language, and advance the PR out of Draft status. I'll be willing to approve it, even if we stick with the same generic svg for all of the languages at this stage.

PR #829, opened 2022-01-14
Merge 1.4.1 release changes into prerelease branch

gjsjohnmurray:
Earlier I created a new branch called **prerelease** from the source that I erroneously published manually to Marketplace today as 1.3.2022011401 after a `vsce package` command that omitted the `--pre-release` switch.

This PR updates that branch with the fixes and enhancements we released today as 1.4.1

The prerelease branch is intended to be long-living. In future it'd be good to add CI so we can publish pre-releases more automatically from this, which should ensure the `--pre-release` switch doesn't get forgotten again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## [1.4.1] 14-Jan-2022
+- Fixes
+  - Version 1.4.0 is failing to activate (#827)
+
 ## [1.4.0] 14-Jan-2022
 - Enhancements
   - Make `Ctrl / Cmd+T` lookup (Open Symbol by Name) check all servers connected to a multi-root workspace (#815)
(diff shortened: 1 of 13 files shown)


isc-bsaviano (approved):
I like the idea of having a separate pre-release branch

isc-bsaviano:
@gjsjohnmurray How does this code behave if someone isn't using the server manager? I know of a few InterSystems devs who work client-side and don't use the server manager because they only ever work on one server at once. I thiink that if the user specifies the password in plain text in `objectscript.conn` then it's reasonable for this extension to expose it in its exported API

gjsjohnmurray:
@isc-bsaviano AFAIK, at the moment the new security model means that the `serverForUri` API never returns a password, regardless of where this extension may have obtained one.

I can propose a relaxation to this, but I don't want it to be part of this PR which is only intended to update the prerelease branch with what we shipped with 1.4.1.

Please re-approve this (my recent push of package-lock.json dismissed your previous review). Then I'll merge this and open a separate PR.

isc-bsaviano (approved):

PR #831, opened 2022-01-17
Colorize VSCode ObjectScript Output

a-boertien:
This PR adds functionality to colorize the VSCode ObjectScript output window, including displaying different colors for:
- Successful compilation
- Errored compilation
- Individual errors ('ERROR:', etc. prefixes)
- ObjectScript authorization error
- Classes/routines/file names in compilation status
- Quoted strings, numbers, dates/times

Examples:
colorize-output-1
colorize-output-2
colorize-output-3
colorize-output-4
colorize-output-5

### DIFF
--- a/package.json
+++ b/package.json
@@ -490,6 +490,11 @@
         "language": "objectscript-csp",
         "scopeName": "source.objectscript_csp",
         "path": "syntaxes/objectscript-csp.tmLanguage.json"
+      },
+      {
+        "language": "vscode-objectscript-output",
+        "scopeName": "source.vscode_objectscript_output",
+        "path": "syntaxes/vscode-objectscript-output.tmLanguage.json"
       }
     ],
     "snippets": [
--- a/package.json
+++ b/package.json
@@ -450,10 +450,7 @@
       },
       {
         "id": "vscode-objectscript-output",
-        "aliases": [],
-        "mimetypes": [
-          "text/x-vscode-objectscript-output"
-        ]
+        "aliases": []
       }
     ],
     "grammars": [
--- a/package.json
+++ b/package.json
@@ -52,7 +52,8 @@
   "enableProposedApi": true,
   "enabledApiProposals": [
     "fileSearchProvider",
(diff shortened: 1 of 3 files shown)


a-boertien:
Note this is intended as a decent starting point
Possible improvements:
- Handle other permutations of errors
- Handle other warnings/text returned by IRIS that I am not aware of
- Handle languages other than English? (-- does IRIS compilation output change based on language settings?)
- Handle ObjectScript within an error/'TEXT:' output (at the moment, only string/numeric/etc. rules apply: which highlights incorrectly in some circumstances)

daimor:
Could you add a screenshot with the result?

a-boertien:
Screenshots added to description

isc-bsaviano:
I just tested the dev vsix and it adds coloring for every Output channel, not just the ObjectScript one.

gjsjohnmurray:
@isc-bsaviano is correct. Until https://github.com/microsoft/vscode/issues/19561 gets implemented (or something like it) there's no way to set per-channel grammars.

a-boertien:
Thanks, I was not aware. That's unfortunate and understandably not acceptable.

I will close this PR and it can be revisited once microsoft/vscode#19561 is resolved, or another workaround is found.

gjsjohnmurray:
I submitted PR https://github.com/microsoft/vscode/pull/140876 that will allow us to use this when/if accepted.

a-boertien:
@gjsjohnmurray I had a crack at it.. not sure if it's the best way

package.json
```diff
@@ -451,7 +452,7 @@
         "id": "vscode-objectscript-output",
         "aliases": [],
         "mimetypes": [
```
gjsjohnmurray, line 454: I haven't tried this, but I think it should be OK to remove the `mimetypes` array property completely now.

gjsjohnmurray (changes requested):
Looking good, but I think the `mimetypes` array property is now redundant.

PR #852, opened 2022-02-08
Improve `TextSearchProvider`

### LINKED ISSUE #849: `TextSearchProvider` unnecessarily opens many documents
The `TextSearchProvider` opens every document that it gets a match in, which causes the Language Server to parse the text are report diagnostics for each one. This causes a huge drop in LS performance until the search is completed and leads to a messy UI  were the user is prompted with problems for documents that aren't open in any editor tabs. I will open a PR that instead calls `getDoc()` directly, and only for classes because it's not needed for other files.

Screen Shot 2022-02-04 at 7 53 21 AM

### LINKED ISSUE #856: Server side search does not adhere to filters present in the workspace configuration
I have the following workspace configuration:

```
{
	"name": "wip22v1:HSLIB",
	"uri": "isfs://wip22v1:hslib/?filter=*.cls,*.inc,*.MODULE,*.LOC&system=1"
}
```

Although I see % classes in my search (looks like the system flag is adhered to), I do not see MODULE or LOC custom documents.

I have proposed APIs enabled as I see the following in my output channel:
intersystems-community.vscode-objectscript version 1.4.2-beta.3 activating with proposed APIs available.

VS Code version:
Version: 1.64.1
Commit: d6ee99e4c045a6716e5c653d7da8e9ae6f5a8b03
Date: 2022-02-07T17:26:08.977Z
Electron: 13.5.2
Chromium: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Darwin x64 20.6.0

isc-bsaviano:
This PR fixes #849 and fixes #856

Changes made:
- Use `AtelierAPI.getDoc()` instead of `vscode.workspace.openTextDocument()` to get the document contents.
- Limit the number of documents being processed at once so the server doesn’t get overloaded.
- Check the `CancellationToken` and stop if the search was cancelled.
- Return correct line when `multilineMethodArgs` is set.
- Add better support for resolving a line from all class member/attribute combinations.
- Support multiple matches on a single line.
- Use the `filter` and `type` isfs query parameters to avoid searching files that the user doesn’t have in their folder.
- Return useful warning messages and properly log errors.
- Fix the type definition for `SearchMatch`.

### DIFF
--- a/src/api/atelier.d.ts
+++ b/src/api/atelier.d.ts
@@ -45,7 +45,7 @@ export interface ServerInfo {
 
 export interface SearchMatch {
   text: string;
-  line?: number;
+  line?: string | number;
   member?: string;
   attr?: string;
   attrline?: number;
--- a/src/providers/FileSystemProvider/TextSearchProvider.ts
+++ b/src/providers/FileSystemProvider/TextSearchProvider.ts
@@ -2,7 +2,23 @@ import * as vscode from "vscode";
 import { SearchResult, SearchMatch } from "../../api/atelier";
 import { AtelierAPI } from "../../api";
 import { DocumentContentProvider } from "../DocumentContentProvider";
-import { outputChannel } from "../../utils";
+import { notNull, outputChannel, throttleRequests } from "../../utils";
+import { config } from "../../extension";
+import { fileSpecFromURI } from "../../utils/FileProviderUtil";
+
+/**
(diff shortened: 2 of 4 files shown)


gjsjohnmurray (approved):
Happy to approve this based on code-reading. I haven't tested it though.

PR #862, opened 2022-02-09
Have `FileSearchProvider` use `DocumentContentProvider.getUri()` to generate URIs

### LINKED ISSUE #861: When using Quick Open, can have multiple version of the same document open in separate tabs
For certain classes (only reproduced with % classes when using system=1 in a non-%SYS namespace), if I open the class using Quick Open, and then open it from the file explorer, I can get two tabs both with the same document open as can be seen in the below screenshots:

Screen Shot 2022-02-09 at 1 31 05 PM
Screen Shot 2022-02-09 at 1 31 13 PM

Also note that in this case, when opening with Quick Open, VS code doesn't take me to where the document exists in the file explorer (does when I switch to a tab opened via the file explorer).

Extension version: 1.4.2-beta.6

VS Code version:
Version: 1.64.1
Commit: d6ee99e4c045a6716e5c653d7da8e9ae6f5a8b03
Date: 2022-02-07T17:26:08.977Z
Electron: 13.5.2
Chromium: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Darwin x64 20.6.0

isc-bsaviano:
This PR fixes #861

### DIFF
--- a/src/providers/FileSystemProvider/FileSearchProvider.ts
+++ b/src/providers/FileSystemProvider/FileSearchProvider.ts
@@ -1,8 +1,9 @@
 import * as vscode from "vscode";
 import * as url from "url";
-import { AtelierAPI } from "../../api";
 import { StudioOpenDialog } from "../../queries";
 import { studioOpenDialogFromURI } from "../../utils/FileProviderUtil";
+import { notNull } from "../../utils";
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (approved):
LGTM

PR #865, opened 2022-02-12
Add `objectscript-int` language id for .int files (#823)

### LINKED ISSUE #823: MAC syntax is allowed in INTs
At the moment we treat INTs and MACs as the same language. IMO this is a mistake, because it means I can write in my INTs code which is only valid in MACs (e.g. #-directives, embedded SQL etc) and only find out when I try to save/import them into a namespace.

To fix this I think we should add a new language `objectscript-int` and associate the .int extension with this rather than with `objectscript`.

_Originally posted by @gjsjohnmurray in https://github.com/intersystems-community/vscode-objectscript/issues/822#issuecomment-1012949887_

@isc-bsaviano - Work will also be needed in the [Language Server](https://github.com/intersystems/language-server).

gjsjohnmurray:
This PR fixes #823

Depends on https://github.com/intersystems/language-server/issues/217 so PR is being created as a draft.

When this extension falls back to TM grammar in the absence of LS (e.g. on web, currently) MAC syntax continues to be allowed in INTs because the new language id still uses the objectscript.tmLanguage.json file

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "onCommand:vscode-objectscript.hideExplorerForWorkspace",
     "onCommand:vscode-objectscript.showExplorerForWorkspace",
     "onLanguage:objectscript",
+    "onLanguage:objectscript-int",
     "onLanguage:objectscript-class",
     "onLanguage:objectscript-macros",
     "onLanguage:xml",
@@ -416,7 +417,15 @@
           "ObjectScript"
         ],
         "extensions": [
-          ".mac",
+          ".mac"
+        ]
+      },
+      {
+        "id": "objectscript-int",
+        "aliases": [
+          "ObjectScript INT"
+        ],
+        "extensions": [
           ".int"
         ]
       },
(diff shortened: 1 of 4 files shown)


gjsjohnmurray:
Ready for review now that LS 1.2.9 has been published.

isc-bsaviano:
@gjsjohnmurray With this dev vsix installed and the LS disabled #define is still allowed in an `.int` file. The #define line is colored correctly and I would expect some type of error to be shown like the LS does.

gjsjohnmurray:
@isc-bsaviano that's what I mean in this sentence of the PR description:

> When this extension falls back to TM grammar in the absence of LS (e.g. on web, currently) MAC syntax continues to be allowed in INTs because the new language id still uses the objectscript.tmLanguage.json file

People are no worse off than they already are. In future maybe someone (@daimor ?) could carve out the subset of the TM grammar which INTs should allow.

isc-bsaviano (approved):

PR #867, opened 2022-02-18
Generate content when a new local class or routine is created

### LINKED ISSUE #864: Documentation enhancement request: Routine creation and compilation
When creating routine files and add .mac/.int in the file name under **Explorer view** and **ObjectScript view** via "New File" utility, it generates an empty file without "ROUTINE <routine name>" declaration on top. Users need to manually add this header otherwise compiling the routine would fail. Creating routine files in a **workspace** gives different behavior, where the ROUTINE <routine name> header is auto generated when adding .mac/.int in the file name. A number of members in InterSystems WRC Support team have noticed this inconsistency and we feel this can bring confusion. We believe this should be documented anywhere, or we may want to change the behavior to be consistent where the ROUTINE header is always generated under all views.

isc-bsaviano:
This PR fixes #864

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -18,6 +18,7 @@ export const schemas = [
 export const filesystemSchemas = [FILESYSTEM_SCHEMA, FILESYSTEM_READONLY_SCHEMA];
 
 import * as url from "url";
+import path = require("path");
 import {
   importAndCompile,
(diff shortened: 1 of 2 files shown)


daimor (approved):

gjsjohnmurray (approved):
Based on code review this all looks fine.

PR #881, opened 2022-02-25
Fix regression that broke server-side editing from ObjectScript Explorer

### LINKED ISSUE #879: Cannot Open Server-side Code Since 1.4.2
Issue Type: <b>Bug</b>

Since installing 1.4.2 when trying to open a serverside copy of the class using the ObjectScript Explorer, I consistently get a messages informing me that, "The editior could not be opened because the file was not found". Copying the file to my local workspace allows me to open it though.

Extension version: 1.4.2
VS Code version: Code 1.64.2 (f80445acd5a3dadef24aa209168452a3d97cc326, 2022-02-09T22:00:58.347Z)
OS version: Darwin x64 20.6.0
Restricted Mode: No

gjsjohnmurray:
This PR fixes #879

The recent work to support copying in isfs folders in VS Code Explorer broke ObjectScript Explorer's ability to open server-side files for editing when configured to do this.

### DIFF
--- a/src/providers/FileSystemProvider/FileSystemProvider.ts
+++ b/src/providers/FileSystemProvider/FileSystemProvider.ts
@@ -369,7 +369,9 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
             throw vscode.FileSystemError.FileNotFound(uri);
           }
           // Caller granted us permission to create structures for intermediate directories not yet seen.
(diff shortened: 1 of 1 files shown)


gjsjohnmurray:
Problem occurred when ObjectScript Explorer uses isfs to enable server-side editing, and also when reloading a workspace with isfs-type folders from which isfs documents were open when the workspace was previously closed.

isc-bsaviano (approved):
This worked for me but I think we should wait for one of the issue raisers to confirm as well, then do a new release when this is merged.

PR #882, opened 2022-02-25
Add `objectscript.openClassContracted` setting

### LINKED ISSUE #876: Optionally fold all ranges when class is opened
> In studio there was the option to "Open class in contracted view"
In vs code, for each class I open, I have to click in the class and ctrl+k ctrl+1 to get the same effect.
Can it be made a setting and auto-applied?

This is possible using the `editor.foldAll` built  in command, which  s what gets called when you use the `ctrl-k-0` keybinding.

*Moved from https://github.com/intersystems/language-server/issues/218*

isc-bsaviano:
This PR fixes #876 

When `objectscript.openClassContracted` is `true`, automatically collapse all folding ranges when a class is opened for the first time.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -738,7 +738,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
     }
   }
 
-  // The URI's of all classes that have been opened. Used when objectscript.openClassContracted is true.
+  // The URIs of all classes that have been opened. Used when objectscript.openClassContracted is true.
   const openedClasses: string[] = [];
 
   context.subscriptions.push(
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -738,6 +738,9 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
     }
   }
 
+  // The URI's of all classes that have been opened. Used when objectscript.openClassContracted is true.
+  const openedClasses: string[] = [];
+
   context.subscriptions.push(
     reporter,
     panel,
@@ -968,6 +971,17 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
           })
       )
     ),
+    vscode.window.onDidChangeActiveTextEditor((editor: vscode.TextEditor) => {
+      if (
+        config("openClassContracted") &&
+        editor &&
+        editor.document.languageId === "objectscript-class" &&
+        !openedClasses.includes(editor.document.uri.toString())
+      ) {
+        openedClasses.push(editor.document.uri.toString());
+        vscode.commands.executeCommand("editor.foldAll");
+      }
+    }),
 
     /* Anything we use from the VS Code proposed API */
     ...proposed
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -738,6 +738,9 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
     }
   }
 
+  // The URI's of all classes that have been opened. Used when objectscript.openClassContracted is true.
```
gjsjohnmurray, line 741: Nitpick - remove apostrophe:

```suggestion
  // The URIs of all classes that have been opened. Used when objectscript.openClassContracted is true.
```

gjsjohnmurray (commented):
What happens when reopening after hot exit? My guess is the open classes all get collapsed.

isc-bsaviano:
@gjsjohnmurray I would expect that open classes would be collapsed when they become the active editor but I haven't tested it because I don't have #881 in that branch.

gjsjohnmurray:
@isc-bsaviano once #881 merges and you update your branch I think it's worth investigating. Maybe move the `openedClasses` array into `workspaceState` so it survives reloads.

isc-bsaviano:
@gjsjohnmurray If we persist `openClasses` then I think we should also remove classes from it when they've been closed.

gjsjohnmurray:
Should perhaps do that even if we don't persist.

isc-bsaviano:
@gjsjohnmurray I confirmed that when restoring from a hot exit, the class is auto-folded again even though it was open and some ranges were left unfolded. I will store `openClasses` in `workspaceState` so this doesn't happen.

gjsjohnmurray (commented):
May be less of an overhead if you fetched `openedClasses` from workspace storage at the start of `activate`, updated the local array on the callbacks, then stored it during `deactivate`.

src/extension.ts
```diff
@@ -969,7 +974,6 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
       )
     ),
     vscode.window.onDidChangeActiveTextEditor((editor: vscode.TextEditor) => {
```
gjsjohnmurray, line 976: Suggest you use a `const` for `editor.document.uri.toString()` in place of the two occurrences below.

src/extension.ts
```diff
@@ -978,15 +982,12 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
       ) {
         vscode.commands.executeCommand("editor.foldAll");
         openedClasses.push(editor.document.uri.toString());
-        workspaceState.update("openedClasses", openedClasses);
       }
     }),
     vscode.workspace.onDidCloseTextDocument((doc: vscode.TextDocument) => {
```
gjsjohnmurray, line 987: Suggest you use a `const` for `doc.uri.toString()` in place of the two occurrences below.

gjsjohnmurray (changes requested):
A couple of improvements suggested.

gjsjohnmurray (approved):

PR #888, opened 2022-03-01
Add `Reporting Issues` page to docs

### LINKED ISSUE #874: Add Feedback instructions to  online documentation
In #462 these were added to the README. I suggest they also get added to the doc at https://intersystems-community.github.io/vscode-objectscript/

isc-bsaviano:
This PR fixes #874 

I just copied the info from the README into a new page on the docs website. Should I remove it from the README as well so we don't need to maintain two copies? I'm indifferent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 
 - Documentation on [GitHub Pages](https://intersystems-community.github.io/vscode-objectscript/).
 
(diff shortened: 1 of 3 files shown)


isc-rsingh:
IMHO we should remove this from the README so that we don't need to maintain 2 versions, and also so that the README gets shorter. It's too long as it is.

gjsjohnmurray (approved):

PR #890, opened 2022-03-02
Compile asynchronously

isc-bsaviano:
This PR removes the use of the synchronous `/action/compile` API and instead uses the `/work` endpoints to compile documents. We've had reports internally of users running into `504 Gateway Timeout` errors when compiling a class took more than 60 seconds. While the users could increase the timeout, they could always run up against it again in the future. Using this approach ensures that the extension will not break, regardless of how long compilation on the server may take.

I'm leaving this as a draft until the people who ran into the gateway timeout issue internally verify that this works for them.

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -246,6 +246,7 @@ export class AtelierAPI {
     if (minVersion > apiVersion) {
       return Promise.reject(`${path} not supported by API version ${apiVersion}`);
     }
+    const originalPath = path;
     if (minVersion && minVersion > 0) {
       path = `v${apiVersion}/${path}`;
     }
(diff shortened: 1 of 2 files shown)


daimor (dismissed):

isc-bsaviano:
@daimor @gjsjohnmurray The internal devs verified that this fixed the 504 error. I think you should both try this out before it gets merged though.

gjsjohnmurray:
Maybe cancellation should add a message to the output channel, in case the partially-compiled classes on the server result in unexpected behaviour.

gjsjohnmurray:
@isc-bsaviano as mentioned on today's call, you'll look to address my feedback re a cancel message. I don't see this as a showstopper though, so if @daimor is happy with what's already in this PR but you don't get to the cancel message thing imminently I suggest merging as-is so it can get into the upcoming release.

gjsjohnmurray (dismissed):

isc-bsaviano:
@gjsjohnmurray I will add a cancellation message, I think that's good to have

isc-bsaviano:
@gjsjohnmurray If the message looks ok to you I will merge this.

gjsjohnmurray (approved):
:+1:

PR #892, opened 2022-03-03
Don't log `Non-JSON response` errors to the Output channel

### LINKED ISSUE #853: /api/atelier/v4 endpoint issue with UCR 2021.2 SSO?
Generally, extension works fine with latest UCR release 2021.2 that has SSO implemented using ISC's OAuth 2.0 model. There's no explicit instructions provided in ISC Docs (or here) on how to adjust the /api/atelier web application to properly work with the new SSO model but interpreting from other areas but haven't run into any issues until today.

When trying to modify an existing class and then compiling (ISFS is off), it is trying to suggest the Server Side copy is newer and normally would see a prompt to override this. Instead, in the output window I see:

```
Non-JSON response to /api/atelier/v4/MYNS/doc/My.Modified.cls?ignoreConflict=0
The page was not displayed because there was a conflict.
```

Audit log in IRIS shows the following

```
Error message: ERROR #862: User UnknownUser is restricted from running application /api/atelier, %Development:U required -- cannot execute.Web Application: /api/atelier
```

So clear it's attempting to execute the original API call without context of the user I'm currently running as is but unclear why this is since all other functionality works as my user (has %All) including the language server extension pieces. 

If I add objectscript.overwriteServerChanges = true to my VSC config, I can bypass this issue and successfully compile but I'd prefer not to leave in this state, especially for my other users. Any guidance on adjusting /api/atelier web app settings to work kindly with UCR 2021.2 SSO OAuth2 model?

isc-bsaviano:
This PR fixes #853 

I modified the error handling so that `Non-JSON response` errors are treated the same as any other. This will eliminate the many non-helpful error messages that flood the Output channel.

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -327,7 +327,8 @@ export class AtelierAPI {
         // User likely ran out of licenses
         throw {
           statusCode: response.status,
-          message: `The server at ${host}:${port} is unavailable. Check License Usage.`,
+          message: response.statusText,
+          errorText: `The server at ${host}:${port} is unavailable. Check License Usage.`,
         };
       }
       if (response.status === 401) {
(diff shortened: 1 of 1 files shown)


src/api/index.ts
```diff
@@ -355,16 +356,17 @@ export class AtelierAPI {
...
+      let data: Atelier.Response;
+      try {
+        data = JSON.parse(responseString);
+      } catch {
         throw {
-          statusCode: 500,
-          message: `Non-JSON response to ${path} request. View 'ObjectScript' channel on OUTPUT tab of Panel for details.`,
+          statusCode: response.status,
+          message: response.statusText,
+          errorText: `Non-JSON response to ${path} request.`,
```
gjsjohnmurray, line 366:
```suggestion
          errorText: `Non-JSON response to ${path} request. Is the web server suppressing detailed errors?`,
```

gjsjohnmurray (changes requested):
One suggestion made. Also not tested by me.

gjsjohnmurray (approved):

PR #896, opened 2022-03-04
Fix `TextSearchProvider` for csp folders

isc-bsaviano:
Previously the `TextSearchProvider` wouldn't work for csp folders because the generated URIs were malformed. This PR fixes that issue and changed the StudioOpenDialog file spec for the root csp folder from `*` to `*.cspall` so only files in csp directories are searched.

### DIFF
--- a/src/providers/DocumentContentProvider.ts
+++ b/src/providers/DocumentContentProvider.ts
@@ -70,7 +70,6 @@ export class DocumentContentProvider implements vscode.TextDocumentContentProvid
           query: "",
         });
       }
-      return uri;
     } else {
       const conn = config("conn", workspaceFolder);
       const localFile = this.getAsFile(name, workspaceFolder);
@@ -132,8 +131,9 @@ export class DocumentContentProvider implements vscode.TextDocumentContentProvid
         });
       }
     }
+    const params = new URLSearchParams(uri.query);
     if (namespace && namespace !== "") {
-      if (isCsp) {
+      if (isCsp && !params.has("csp")) {
         uri = uri.with({
           query: `ns=${namespace}&csp=1`,
         });
@@ -142,7 +142,7 @@ export class DocumentContentProvider implements vscode.TextDocumentContentProvid
           query: `ns=${namespace}`,
         });
       }
-    } else if (isCsp) {
+    } else if (isCsp && !params.has("csp")) {
       uri = uri.with({
         query: "csp=1",
       });
--- a/src/providers/DocumentContentProvider.ts
+++ b/src/providers/DocumentContentProvider.ts
@@ -60,7 +60,7 @@ export class DocumentContentProvider implements vscode.TextDocumentContentProvid
         .join(/cls|mac|int|inc/i.test(fileExt) ? "/" : ".");
       name = fileName + "." + fileExt;
       uri = wFolderUri.with({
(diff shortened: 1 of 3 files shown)


src/providers/DocumentContentProvider.ts
```diff
@@ -70,6 +70,7 @@ export class DocumentContentProvider implements vscode.TextDocumentContentProvid
           query: "",
         });
       }
+      return uri;
```
gjsjohnmurray, line 73: @isc-bsaviano Returning early here will cause subsequent code (line 134+) to be bypassed. Are you certain that it will never be necessary to add the ns and/or csp queryparams? I think a user's isfs root folder spec can still use ?ns=USER instead of the preferred isfs://servername:user/ syntax.

gjsjohnmurray (changes requested):
See my comment about how this seems no longer to append the `ns` and `csp` queryparams in some cases where it may still be necessary.

src/providers/DocumentContentProvider.ts
isc-bsaviano: @gjsjohnmurray I'm not sure, I need to look this over again and see why I made that change.

isc-bsaviano: @gjsjohnmurray I added that return there because the code on line 145 replaces the existing `csp` query param with `csp=1`, leading to a different URI than the file in the tree. I will remove the return statement and will condition the code on line 145 so it only modifies the URI if the csp param isn't present at all

src/providers/DocumentContentProvider.ts
```diff
@@ -132,8 +131,9 @@ export class DocumentContentProvider implements vscode.TextDocumentContentProvid
         });
       }
     }
+    const params = new URLSearchParams(uri.query);
     if (namespace && namespace !== "") {
-      if (isCsp) {
+      if (isCsp && !params.has("csp")) {
```
gjsjohnmurray, line 136: Is the extra condition necessary? Won't it also mean that if wFolderUri already specified `?csp=1` (say) and a value for `namespace` was passed, the uri returned by this function will lack the `csp=1` queryparam?

src/providers/DocumentContentProvider.ts
```diff
@@ -142,7 +142,7 @@ export class DocumentContentProvider implements vscode.TextDocumentContentProvid
           query: `ns=${namespace}`,
         });
       }
-    } else if (isCsp) {
+    } else if (isCsp && !params.has("csp")) {
```
gjsjohnmurray, line 145: Also seems unnecessary as above, though in this case maybe harmless.

gjsjohnmurray (changes requested):

src/providers/DocumentContentProvider.ts
isc-bsaviano: This is the one that's necessary for this PR. Otherwise the generated URI won't match the file system and the user could potentially open two copies of the document at the same time.

src/providers/DocumentContentProvider.ts
isc-bsaviano: Yes, you're correct. I'll fix it

gjsjohnmurray (approved):

PR #905, opened 2022-03-24
Improve `objectscript.openClassContracted` setting

isc-bsaviano:
See https://github.com/intersystems-community/vscode-objectscript/issues/876#issuecomment-1077333926


gjsjohnmurray:
Some people might prefer the complete collapse. We could allow more values on the setting. Or maybe make this change and wait to see if anyone complains.

isc-bsaviano:
@gjsjohnmurray foldLevel1 is more in line with what Studio does. Since the purpose of adding this is feature parity with Studio, I think it should match. Me using foldAll initially was a mistake IMO.

gjsjohnmurray (approved):

PR #910, opened 2022-03-31
Finalize `outputChannelLanguage` API

isc-bsaviano:
VS Code 1.66 was released and with it the `outputChannelLanguage` API was finalized. We should use the new finalized API so everyone will get their Output channel colorized.


gjsjohnmurray (approved):
LGTM

PR #917, opened 2022-04-04
Add `objectscript.export.exactFilter` setting

### LINKED ISSUE #913: Export Filter Enhancement
I am using the `"objectscript.export.filter"` setting in my settings.json
```
    "objectscript.export.filter": "CH.%.cls",
```
This works similar to the filter in the export UI, but is rather limited. I would like to limit the export to just the names that start with `CH.` using a filter value of `CH.%.cls`, but per the documentation there is an implicit wildcard before and after the filter criteria:
> The filter is applied to document names using the [LIKE predicate](https://irisdocs.intersystems.com/irislatest/csp/docbook/DocBook.UI.Page.cls?KEY=RSQL_like) (i.e. Name LIKE '%filter%').

This results in hundreds of matches for us where "ch.%.cls" is found somewhere else after the beginning of the name. 
image

I am working around this by manually filtering the returned results and unselecting the undesired results, but it would be ideal if we could just get the results we are looking for with configuration.

isc-bsaviano:
This PR fixes #913


gjsjohnmurray (approved):
Haven't tested, but it looks good.

PR #918, opened 2022-04-04
Fix importing of web app files

### LINKED ISSUE #777: CSP and JS cannot be synchronized to the server
When the code is checked out locally, the compiled CSP and JS cannot be synchronized to the server

isc-bsaviano:
This PR fixes #777

Whenever the connection is checked I cache the list of available web apps in a server-namespace and use this list to determine if a local non-CSP/CSR file can be imported as a web app file.


gjsjohnmurray (approved):
LGTM

PR #920, opened 2022-04-06
Prepare 1.6.0 release

gjsjohnmurray:


isc-bsaviano:
@gjsjohnmurray What work is left to do before we can merge the authentication API stuff and junk the pre-releases? A few devs internally have gotten into a mess by installing pre-releases without knowing what they are so I would like to get rid of those soon if possible. If the authentication API work is close to being done I propose that we hold off on 1.6.0 until then. If not we just jump right to 1.8.0 when it's ready.

gjsjohnmurray:
@isc-bsaviano still to be done in Server Manager are:

- [Migrate Server Manager 2 stored passwords](https://github.com/intersystems-community/intersystems-servermanager/issues/129)
- [Update Windows registry importer to target new Server Manager 3 secure password storage](https://github.com/intersystems-community/intersystems-servermanager/issues/130)

It's likely to be a while until I have bandwidth to do these.

isc-bsaviano:
Thanks for posting those issues so I know what still needs to be done. Let's just release 1.6.0 and worry about that later. All of the work for this extension is done, correct? The LS support has been out for a while now.

isc-bsaviano (approved):

PR #934, opened 2022-04-20
Prepare 1.6.1 release

gjsjohnmurray:


isc-bsaviano:
@gjsjohnmurray I think this should be a 1.8.0 release because it contains a significant enhancement (projects support).

PR #949, opened 2022-05-31
add some settings doc

isc-rsingh:
Add some documentation of objectscript.conn settings, including the very useful docker-compose object


isc-rsingh:
Also reduces the left margin on all pages and makes the first column of the InterSystems Objectscript table narrower so that at least some of the Notes column should show on even small windows. Previously it was likely that no one would ever know that column existed.

isc-bsaviano (approved):

PR #959, opened 2022-07-22
Use `DocumentSymbolProvider` for tag+offset status bar item

### LINKED ISSUE #957: Incorrect tag+offset^routine display in status bar
We very nicely display the label+offset^routine in the VS Code bottom status bar, so that you find a specified line in the source.  However, we get this wrong with tags inside procedure block.  Tags inside a procedure block should not be picked up as a label for the label+offset display as they are not valid labels. 

*Reported internally*

isc-bsaviano:
This PR fixes #957 

Call the `vscode.executeDocumentSymbolProvider` command and use the symbols to determine label+offset for the status bar item.

***NOTE:*** This PR won't fix the issue until the next Language Server version is released due to a dependency on chnages to its `DocumentSymbolProvider`.


gjsjohnmurray (approved):
LGTM

PR #961, opened 2022-07-25
Use open document's server for `attach` debug sessions

### LINKED ISSUE #958: Debugger 'Attach' using isfs
Our development team relies heavily on editing files with isfs, so I have only observed this when using isfs.  When trying to use the debugger to attach to a process, I am unable to get the list of processes to populate.  This appears to be because the handler for the 'vscode-objectscript.pickProcess' command creates a new AtlierAPI with no arguments.  Inside the APIs constructor, this causes the current folder via currentWorkspaceFolder to be used for setConnection.  If the first folder's "name" happens to be the same as one of the defined intersystems.servers entries, this will work correctly.  However, if using the default folder name of server:namespace, the connection is not set correctly, so the API call to getJobs will always fail.  The workaround is to simply rename the folder to be the same as the current server, but this is not ideal.  

This can be fixed by initializing the API with vscode.window.activeTextEditor.document.uri as the first argument.

This allows the API to correctly retrieve the process list from the server associated with the currently open document.  This introduces a requirement to have a file open, but seems to be in line with the following from the documentation:

"When starting objectscript launch debug session, make sure that the file containing the program that you are debugging is open in your editor and is the active tab."

isc-bsaviano:
This PR fixes #958

### DIFF
--- a/docs/RunDebug.md
+++ b/docs/RunDebug.md
@@ -108,7 +108,7 @@ You can select a debug configuration from the list VS Code provides in the Run a
 
 Clicking on the green arrow runs the currently selected debug configuration.
 
-When starting **objectscript launch** debug session, make sure that the file containing the **program** that you are debugging is open in your editor and is the active tab. VS Code will start a debug session with the server of the file in the active editor (the tab that the user is focused on).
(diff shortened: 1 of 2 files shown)


isc-bsaviano:
@gjsjohnmurray Once this PR gets merged can you do the next release? I don't have anything else to pull in.

PR #972, opened 2022-08-10
Reduce VSIX size

isc-bsaviano:
The VSIX for this extension is incredibly large (6.2MB for 1.8.3-beta.1). I've updated the `.vscodeignore` to more aggressively exclude unneeded files. A SNAPSHOT version of the extension built on my mac with this change was just 1.5MB, a much more acceptable number.

@gjsjohnmurray I ran `vsce ls` and tested the extension a bit using my `isfs` workspace and it appears to work fine but I would appreciate some manual review by you before this gets merged because I excluded a lot of new files.


gjsjohnmurray:
@isc-bsaviano I think a simpler solution would be to adopt .vscodeignore from the prerelease branch. The history of the file in that branch will show you how on Nov 26, 2021 I backed out the changes I previously had to make in June of that year when a breaking change landed in vsce@1.92.0

isc-bsaviano:
The pre-release branch is still quite a bit larger than this branch. I will look at its .vscodeignore

gjsjohnmurray:
> The pre-release branch is still quite a bit larger than this branch. I will look at its .vscodeignore

That could be an indication that you've excluded too much.

the `vsce ls` command will simulate a build of the VSIX sufficiently to list the names of all the files in it. I suggest you compare what this gives for the branch of this PR with what you get when you switch to the prerelease branch and run the same command.

isc-bsaviano:
@gjsjohnmurray I'm going to update this using the prerelease branch version as a base because I was missing the webview stuff. Should I include the .map files though? I'm not sure if VS Code uses these for debugging released extensions and since they add about 50% to the bundle size it would be nice to exclude them.

daimor:
.map files should not even appear. They are only for development time

isc-bsaviano:
Updated based on the prerelease branch

.vscodeignore
```diff
@@ -1,16 +1,14 @@
...
-# but vsce@1.92.0 broke it (see https://github.com/microsoft/vscode-vsce/issues/576)
-
-/*
-!/dist
-/dist/*.map
-!/images
-!/snippets
-!/syntaxes
+**
+!dist/*.js
```
gjsjohnmurray, line 2:
```suggestion
!dist/*.js
!dist/*.txt
```
To include a license file that we are probably obliged to ship.

.vscodeignore
```diff
@@ -1,16 +1,14 @@
-# Run 'vsce ls' to verify that the following lines exclude everything not needed
```
gjsjohnmurray, line 1: I suggest we retain the comment line hint about how to check what this file allows into the VSIX

.vscodeignore
```diff
@@ -1,16 +1,14 @@
...
-!/dist
-/dist/*.map
-!/images
-!/snippets
-!/syntaxes
+**
+!dist/*.js
+!snippets/
+!images/*.png
+!images/*.svg
```
gjsjohnmurray, line 5: Can we reduce the two lines above to just this?
```
!images/
```

It looks like installation.gif is the only one that youre 2-line approach excludes, and since that file is apparently no longer used I suggest you delete it as part of this PR.

gjsjohnmurray (changes requested):

gjsjohnmurray (approved):

PR #978, opened 2022-08-29
implementing #974

### LINKED ISSUE #974: Add "Run Method in Terminal" feature
There is a handy "Debug this method" feature above every class method.
Could we also add "Run Method in Terminal" or similar to copy the line suitable to execute classmethod in Terminal?

E.g. if the method is "
Class Foo {
classmethod Bar () As %Status {
}

The function will copy the following:
w ##class(Foo).Bar()

daimor:
This PR fixes #974 

ezgif com-gif-maker (6)


isc-bsaviano:
@daimor I like the addition of the "debug this method" code lens for routines but I'm strongly against the "run this method" code lens for the reasons outlined in my original comment on #974 so I'm not going to approve this.

isc-rsingh (approved):

PR #985, opened 2022-09-22
Add `objectscript.importOnSave` setting

### LINKED ISSUE #982: Feature Request: Add "import (upload) on save" setting
I'm relatively new to using this extension, so I might be overlooking something obvious.

From what I can tell, an exported file is always imported (uploaded) to the server on saving.
To my surprise, this even happens when disabling the "compile on save" option.
This means I cannot easily save local changes privately - unless there happens to be a conflict with the server version and I discard/ignore the corresponding info window.

So basically I'd appreciate the option to deliberately decouple saving and importing (uploading) into separate steps:
Either a new flag, or expand the "compile on save" option to become a drop-down: "on save" -> {"do nothing", "import/upload", "import/upload and compile"}

For now there are probably workarounds, I can just remember never to save it until I want to re-import the code, or move the file out of the workspace (EDIT: renaming it didn't seem to work).
(On that topic, I'm not sure what benefits or use cases even exists for having a frequently-synced exported file over directly using server-side editing,
but if one exists then the current behavior can be kept; I'd just prefer the ability to work locally without overwriting server-side code.)

If the idea sounds good I might find the time to implement it myself within the next month if no-one else is up for it until then.

isc-bsaviano:
This PR fixes #982


gjsjohnmurray (approved):
LGTM but I'm not a big user of client-side editing mode.

PR #988, opened 2022-09-27
Custom background color for Generated

daimor:
The proposed realization for the discussed topic #987

### DIFF
--- a/package.json
+++ b/package.json
@@ -977,6 +977,38 @@
       "title": "InterSystems ObjectScript",
       "type": "object",
       "properties": {
+        "objectscript.generatedCodeBackground": {
+          "description": "Background color of the editor for generated code",
+          "type": "object",
+          "additionalProperties": false,
+          "required": [
+            "enabled"
+          ],
+          "properties": {
+            "enabled": {
+              "type": "boolean",
+              "default": false
+            },
+            "light": {
+              "type": "string",
+              "format": "color-hex",
+              "default": "#ddd"
+            },
+            "dark": {
+              "type": "string",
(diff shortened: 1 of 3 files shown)


gjsjohnmurray (changes requested):
This is ugly because it affects other visible editors, not just the one that contains generated code.

There is a better way using [TextEditorDecorationType](https://code.visualstudio.com/api/references/vscode-api#TextEditorDecorationType)

```ts
	const readonlyDocumentDecorationType = vscode.window.createTextEditorDecorationType({
		isWholeLine: true,
		backgroundColor: new vscode.ThemeColor('objectscript.generatedCodeBackground')
	});
```

Then use [TextEditor](https://code.visualstudio.com/api/references/vscode-api#TextEditor)'s setDecorations method for every line of an editor containing generated code.

For best results use some transparency on the colors you choose.

isc-bsaviano:
@daimor @gjsjohnmurray Is it possible to add an extra icon to the editor tab instead of changing the background color? That would be less intrusive IMO and would remove the need to coordinate colors with themes and possibly introduce accessibility concerns.

gjsjohnmurray:
> Is it possible to add an extra icon to the editor tab

We could experiment with [FileDecorationProvider](https://code.visualstudio.com/api/references/vscode-api#FileDecorationProvider). Perhaps use 'G' as the `badge` property of the `FileDecoration` and 'Generated' as the `tooltip`. Or even add info to the tooltip to say what it is generated from.

isc-bsaviano:
Closing in favor of #1035

PR #991, opened 2022-09-28
Allow `serverForUri` API to return insecure password from `intersystems.servers` (#989)

### LINKED ISSUE #989: Prerelease: Return password from `serverForUri()` if it's specified in `settings.json`

gjsjohnmurray:
This PR fixes #989 

It minimally relaxes a previous constraint in the `serverForUri` API, allowing this function to return the password for a named server if that password is already available to every extension because of having been stored in settings.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1102,14 +1102,14 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
   reporter && reporter.sendTelemetryEvent("extensionActivated");
 
   // The API we export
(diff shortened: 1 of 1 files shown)


gjsjohnmurray:
@isc-bsaviano I think this does what you asked for. Please test the dev VSIX and confirm.

isc-bsaviano (approved):
I tested the dev vsix and it worked as expected. Thanks @gjsjohnmurray

PR #1007, opened 2022-10-27
Improve CodeLenses

### LINKED ISSUE #996: Improve debug/copy CodeLenses
> I'm getting "Debug this Method / Copy Invocation" popping up in inappropriate places... Like labels inside procedure blocks and private procedures.
> 
> Also (a nit) it always calls it a "Method" even when not in a class where it's either a label, a function, or a procedure.

*Caused by #978, reported internally*

isc-bsaviano:
* Fixes #996 
* Adds CodeLenses for INT routines
* Adds documentation for `objectscript.debug.copyToClipboard`
* Shortens debug CodeLens name from `Debug this method` to `Debug`


gjsjohnmurray (approved):
LGTM but I haven't tested it.

PR #1009, opened 2022-10-31
Prepare 2.2.0 release

isc-bsaviano:
See CHANGELOG for what is included.

@gjsjohnmurray Are we sticking with even minor versions even though there are no more prerelease versions?


gjsjohnmurray (approved):
@isc-bsaviano I think we should continue with the even/odd distinction.

isc-rsingh (approved):

PR #1021, opened 2022-11-30
Server-side search: use include/exclude specs (fixes #1019)

### LINKED ISSUE #1019: Errors when trying to use server-side search
Hi,
I'm getting these errors while trying to use the server-side search:
immagine
I've followed these instructions: https://github.com/intersystems-community/vscode-objectscript#enable-proposed-apis
It seems to be installed correctly:
immagine
I'm searching using the ctrl+shift+F combo

_Originally posted by @gbertolottiKS in https://github.com/intersystems-community/vscode-objectscript/issues/126#issuecomment-1331952630_

gjsjohnmurray:
This PR fixes #1019

Unless isfs spec already includes a filter (which it rarely does), apply includes and excludes at the server side when searching using the TextSearchProvider API (still in proposed state).


src/providers/FileSystemProvider/TextSearchProvider.ts
```diff
@@ -104,13 +104,43 @@ export class TextSearchProvider implements vscode.TextSearchProvider {
...
+                ? value.slice(3)
+                : value
+            )
+            .filter((value) => value !== "");
+        };
+        const tidiedIncludes = tidyFilters(options.includes);
+        const tidiedExcludes = tidyFilters(options.excludes);
+        const filter = tidiedIncludes.join(",") + (tidiedExcludes.length === 0 ? "" : ",'" + tidiedExcludes.join(",'"));
+        if (filter) {
+          uri = options.folder.with({ query: `filter=${filter}` });
```
isc-bsaviano, line 134: This is unsafe because it will erase other query parameters. You should hold onto the URLSearchParams object, add the filter to it, then call its `toString()` method to get the value for the uri

src/providers/FileSystemProvider/TextSearchProvider.ts
```diff
@@ -104,13 +104,43 @@ export class TextSearchProvider implements vscode.TextSearchProvider {
     } else {
       const sysStr = params.has("system") && params.get("system").length ? params.get("system") : "0";
       const genStr = params.has("generated") && params.get("generated").length ? params.get("generated") : "0";
+
+      let uri = options.folder;
+
+      if (!new URLSearchParams(uri.query).has("filter")) {
```
isc-bsaviano, line 110: Can we allow this when the filter is present and just append our new filter to the existing one? I'm not sure if that could lead to a malformed filter but it's worth considering.

isc-bsaviano (changes requested):

src/providers/FileSystemProvider/TextSearchProvider.ts
gjsjohnmurray: I had wondered about doing that, but if the isfs spec's filter already lists what to include isn't it the case that adding more includes can never result in less server-side searching, and may unexpectedly cause more? Adding excludes might be OK I suppose.

isc-bsaviano: Yes, I suppose it could lead to more searching if you put something in the includes that isn't in the folder to begin with, like if your folder has `type=cls` but you put `*.mac` in the includes. That would be really dumb but there's nothing stopping you from doing that. Excludes should be fine.

src/providers/FileSystemProvider/TextSearchProvider.ts
```diff
@@ -104,13 +104,46 @@ export class TextSearchProvider implements vscode.TextSearchProvider {
...
+                : value
+            )
+            .filter((value) => value !== "");
+        };
+        const tidiedIncludes = tidyFilters(options.includes);
+        const tidiedExcludes = tidyFilters(options.excludes);
+        const filter = tidiedIncludes.join(",") + (tidiedExcludes.length === 0 ? "" : ",'" + tidiedExcludes.join(",'"));
+        if (filter) {
+          uriQuery.append("filter", filter);
+          uri = options.folder.with({ query: uriQuery.toString() });
```
isc-bsaviano, line 136: This actually could lead to more searching, specifically if the user has a type set that doesn't match what they're including. I.e. `?type=cls` and includes contains `*.mac`.

gjsjohnmurray: I don't think it's practical to try and respect a `?type=xxx` qualifier as well as a "files to include" spec in Search. User needs to be explicit about file extension if they want search results to only be a particular type.

I have pushed a fix for https://github.com/intersystems-community/vscode-objectscript/issues/1019#issuecomment-1333661047

src/providers/FileSystemProvider/TextSearchProvider.ts
```diff
@@ -104,13 +104,51 @@ export class TextSearchProvider implements vscode.TextSearchProvider {
...
+      let uri = options.folder;
+
+      if (!params.get("filter")) {
+        // Unless isfs spec already includes a non-empty filter (which it rarely does), apply includes and excludes at the server side.
+        // If include or exclude field is set to, say, A1B2M*.int there will be two consecutive options.[in|ex]cludes elements:
+        //   **/A1B2M*.int/**
+        //   **/A1B2M*.int
+        //
+        // Ignore first, and strip **/ prefix from second.
+        // When 'Use Exclude Settings and Ignore Files' is enabled (which is typical) options.excludes will also contain entries from files.exclude and search.exclude settings.
```
isc-bsaviano, line 117: Is there any way to ignore this setting or throw away those excludes? I'm testing this now and with the setting on, you get a really long value for `files` that is mostly junk:

```
files=User,User.**,'.git,'.svn,'.hg,'CVS,'.DS_Store,'Thumbs.db,'node_modules,'bower_components,'*.code-search,'*.prj
```

All of these extra file excludes do nothing and since this is all passed in the URL and not the body, we could run into issues with the URL string being too long. My search include is just the User folder in an `isfs` folder with no `type` or `filter`, so the `files` that I'd expect to see is `User.*.cls, User.*.mac,User.*.int,User.*.inc,'*.prj`. You need the 4 extensions because otherwise all file types will be searched, and not all file types are in the tree. Sorry to be nitpicky but this is more complicated than it looks at first glance.

gjsjohnmurray: To strip off the excludes that originate from `files.exclude` and `search.exclude` maybe I can get these settings, dedupe them, count them, then remove that many entries from the end of the excludes array which the search API gives me.

I want to look in more detail at how to handle package stuff.

src/providers/FileSystemProvider/TextSearchProvider.ts
```diff
@@ -104,13 +105,94 @@ export class TextSearchProvider implements vscode.TextSearchProvider {
...
+          const results: string[] = [];
+          filterMap.forEach((_v, key) => {
+            results.push(key);
+          });
+          return results;
+        };
+
+        // Function to get one of the two kinds of exclude settings as an array
+        const getConfigExcludes = (key: string) => {
+          return Object.entries(vscode.workspace.getConfiguration(key, options.folder).get("exclude"))
```
isc-bsaviano, line 156: Is this call to `getConfiguration()` necessary? There's an`options.useIgnoreFiles`,`options.useGlobalIgnoreFiles` and `options.useParentIgnoreFiles` but I'm not sure if they're the same thing

gjsjohnmurray: Yes it's necessary. The three options you mention are only booleans, designed to tell the searchprovider whether or not it should also apply the contents of those files. We're not currently doing anything with them, but I don't see that as a problem at the moment.

isc-bsaviano (approved):
I have one more PR that I want to open but after that let's get a new release out that contains these recent changes.

PR #1029, opened 2022-12-09
Add commands for creating new files

isc-bsaviano:
This PR adds commands for creating new files. They are modeled after the Studio Production new file wizards.


gjsjohnmurray (dismissed):
Nice! I'm happy to approve it as-is, but a possible future enhancement would be to use the multi-step input technique that's demonstrated at https://github.com/microsoft/vscode-extension-samples/blob/main/quickinput-sample/src/multiStepInput.ts

This would allow the user to go back to the previous step if they need to change something.

isc-bsaviano:
@gjsjohnmurray Is multi-step supported when mixing `QuickPick` and `InputBox`? If so I can look into using that next week.

gjsjohnmurray:
> Is multi-step supported

Yes. The example demonstrates such a case.

isc-bsaviano:
Sounds good, I'll check it out.

isc-bsaviano:
@gjsjohnmurray I added support for multi-step prompts except for the production `QuickPick` for rules because it's conditional.

gjsjohnmurray (commented):
My first test was to create a new BS. I could go back from step2 to step 1, but not from step 3 to step 3.

gjsjohnmurray (approved):

PR #1035, opened 2022-12-27
Add `FileDecoration` for Generated files

isc-bsaviano:
This PR fixes #987. I added a "gear" `FileDecoration` for generated files that appears once it has been opened. I also added a setting for turning them off for people (like me) who see this as UI clutter rather than a valuable feature.


gjsjohnmurray (approved):
LGTM but I haven't tested it.

isc-bsaviano:
@gjsjohnmurray Can you install it when you have a minute? I use a gear emoji for the decoration but I can change it to a G if you think people wouldn't be able to tell what it is.

gjsjohnmurray:
Done that now. I think the gearwheel is fine.

PR #1042, opened 2023-01-09
Don't call `openTextDocument` in debugger

isc-bsaviano:
Calling `openTextDocument` causes the Language Server to parse the document and report Diagnostics, which doesn't make sense since the those documents may never be shown in the UI. This PR removes those calls and uses `workspace.fs.readFile()` instead.


gjsjohnmurray (approved):
LGTM

PR #1043, opened 2023-01-10
Add support for debugging REST services

### LINKED ISSUE #521: Debug REST service classes in Visual Studio Code
Is there anything planned for debugging REST service classes in Visual Studio Code like what is done with Serenji : 

img

isc-bsaviano:
This PR fixes #521 

Adds the `Debug REST Service...` command which opens a webview that allows users to configure a REST request. When they are done, the webview will close and the REST debugging session will be started automatically.


daimor:
Could you attach a GIF, of how it works?
Don't get what is WebviewPanel used for

Is not it better to implement through CodeLense injected right to REST Url Map?

isc-bsaviano:
@daimor Here's a GIF. I don't think the `CodeLens` idea would work because there's no way to associate a class with a `UrlMap` to a web application programmatically. We could ask the user for that info but that's not a huge change from the current implementation. 

output

daimor (approved):

PR #1047, opened 2023-01-12
Fix  'No file system provider found' errors when debugging local file

### LINKED ISSUE #1046: 'No file system provider found' when debugging local file
That's for this great features, but it's seems to me it's only works with isfs. Is it possible to make it works with local source file ?

image

error message : No file system provider found for resource 'objectscript://XXX.cls'

_Originally posted by @grongierisc in https://github.com/intersystems-community/vscode-objectscript/issues/521#issuecomment-1379843086_

isc-bsaviano:
This PR fixes #1046 

I implemented option 2 from [this comment](https://github.com/intersystems-community/vscode-objectscript/issues/1046#issuecomment-1380287457). Once this is merged I will release version 2.4.1 with this patch.

### DIFF
--- a/src/debug/debugSession.ts
+++ b/src/debug/debugSession.ts
@@ -18,7 +18,7 @@ import { DebugProtocol } from "vscode-debugprotocol";
 import WebSocket = require("ws");
 import { AtelierAPI } from "../api";
 import * as xdebug from "./xdebugConnection";
-import { schemas } from "../extension";
+import { documentContentProvider, OBJECTSCRIPT_FILE_SCHEMA, schemas } from "../extension";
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (approved):

grongierisc:
I can confirm, this fix the issue :
image

BTW : I think it can be very handy if the Webview can be keep open and display back the response.

Never the less, great job! thanks again for this feature.

isc-bsaviano:
@grongierisc Thanks for confirming. I considered that but I decideed that it would be best if the webview got out of your way during debugging. If you wanted to verify results you should use a full-feature API development tool like Postman. This feature is only meant to be a quick way of starting an ObjectScript debugging session.

PR #1053, opened 2023-01-19
add ? to url if not present before adding token

### LINKED ISSUE #1052: Generates wrong link with serverUrl
https://github.com/intersystems-community/vscode-objectscript/blob/122741f1a7be98ef9ce4e534c69bd8b1a4121fec/src/commands/serverActions.ts#L187

`${serverUrl}/csp/user/menu.csp` will use incorrect URL `/csp/user/menu.csp&CSPCHD=${token}`, insetad of using `?` to split page and query part

daimor:
This PR fixes #1052

### DIFF
--- a/src/commands/serverActions.ts
+++ b/src/commands/serverActions.ts
@@ -184,6 +184,7 @@ export async function serverActions(): Promise<void> {
             const path = vscode.Uri.parse(urlString).path;
             const token = await getCSPToken(api, path);
             if (token.length > 0) {
+              urlString += urlString.includes("?") ? "" : "?";
               urlString += `&CSPCHD=${token}`;
             }
           }
--- a/src/commands/serverActions.ts
+++ b/src/commands/serverActions.ts
@@ -179,16 +179,16 @@ export async function serverActions(): Promise<void> {
           break;
         }
         default: {
-          let urlString = action.detail;
+          let url = vscode.Uri.parse(action.detail);
           if (action.rawLink?.startsWith("${serverUrl}")) {
-            const path = vscode.Uri.parse(urlString).path;
-            const token = await getCSPToken(api, path);
+            const token = await getCSPToken(api, url.path);
             if (token.length > 0) {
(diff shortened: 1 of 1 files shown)


src/commands/serverActions.ts
```diff
@@ -184,6 +184,7 @@ export async function serverActions(): Promise<void> {
             const path = vscode.Uri.parse(urlString).path;
             const token = await getCSPToken(api, path);
             if (token.length > 0) {
+              urlString += urlString.includes("?") ? "" : "?";
```
isc-bsaviano, line 187: It would be safer to check if `urlString.endsWith("?")`

isc-bsaviano: Never mind, that's incorrect.

src/commands/serverActions.ts
```diff
@@ -184,6 +184,7 @@ export async function serverActions(): Promise<void> {
             const path = vscode.Uri.parse(urlString).path;
```
isc-bsaviano, line 184: Instead of manipulating the URL as a string and then parsing it into a `vscode.Uri`, why not parse it from the start and then manipulate it as an object? I think code like this is much better:

```typescript
          let url = vscode.Uri.parse(action.detail);
          if (action.rawLink?.startsWith("${serverUrl}")) {
            const token = await getCSPToken(api, url.path);
            if (token.length > 0) {
              url = url.with({
                query: url.query.length ? `${url.query}&CSPCHD=${token}` : `CSPCHD=${token}`
              });
            }
          }
          vscode.env.openExternal(url);
```

isc-bsaviano:
@gjsjohnmurray How does this look?

gjsjohnmurray:
New code adds CSPSHARE=1 whenever CSPCHD gets added. Perhaps add a comment to this PR explaining why.

isc-bsaviano:
@gjsjohnmurray I think adding that was an oversight. Regardless, it's not relevant to fixing the linked issue so I removed it.

gjsjohnmurray (approved):

PR #1054, opened 2023-01-19
Use query instead of index for class Deployed checks

isc-bsaviano:
The Index API transfers a lot of data that is not needed to check if a class is deployed. It is much more efficient to query `%Dictionary.ClassDefinition` directly.


daimor:
I would use one method for the check and used it instead of calling the same sql query from two places

daimor (approved):

PR #1057, opened 2023-01-25
Use `docker compose` command if present

### LINKED ISSUE #947: Feature Request: Ability to choose docker-compose version
It would be nice to have a configuration option that lets the user customize the docker-compose command being used.

I would like to see configurable settings similar to what the docker extension provides, or perhaps even reusing those configurations if possible. (https://code.visualstudio.com/docs/containers/reference#_command-customization)

I'm having an issue with the extension's use of the docker-compose command that I believe would be resolved if I could update the command to use docker-compose v2.

```Command failed: docker-compose.exe -f conf/docker/docker-compose.yml   ps --services --filter status=running
The Compose file '.\conf/docker/docker-compose.yml' is invalid because:
'name' does not match any of the regexes: '^x-'

You might be seeing this error because you're using the wrong Compose file version. Either specify a supported version (e.g "2.2" or "3.3") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/```

isc-bsaviano:
This PR fixes #947 

I added a function that runs `docker compose version` and if the command doesn't error, `docker compose` is used as the command for compose features instead of the older `docker-compose`.

### DIFF
--- a/src/commands/serverActions.ts
+++ b/src/commands/serverActions.ts
@@ -104,14 +104,14 @@ export async function serverActions(): Promise<void> {
     actions.push({
       id: "openDockerTerminal",
       label: "Open Terminal in Docker",
-      detail: "Use docker-compose to start session inside configured service",
+      detail: "Use Docker Compose to start session inside configured service",
     });
   }
   if (workspaceState.get(workspaceFolder.toLowerCase() + ":docker", false)) {
     actions.push({
       id: "openDockerShell",
       label: "Open Shell in Docker",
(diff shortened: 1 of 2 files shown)


daimor (approved):

gjsjohnmurray (approved):
Looks OK to me, but untested as this isn't a mode I use the extension in.

PR #1062, opened 2023-01-29
Trigger reload of any `objectscript://` document after import

gjsjohnmurray:
This PR ensures that if a class/routine/etc was loaded read-only by another extension using an `objectscript://` uri, then if the user imports an updated copy of the same file (client-side or server-side development paradigm) the read-only document reloads.


gjsjohnmurray:
@isc-bsaviano are you able to review and approve this? I'd like it merged so a new extension of mine can benefit from it.

isc-bsaviano:
@gjsjohnmurray Sorry I have been on PTO. What happens when the document isn't open and this event fires? Does that have any negative side effects?

gjsjohnmurray:
@isc-bsaviano no side effects that I've seen.

isc-bsaviano (approved):

PR #1066, opened 2023-02-01
Added command to "Refresh file contents"

ollitanska:
This PR will add a new command "Refresh file contents". This command will be shown alongside with the "Import and Compile" entry in the VSCode explorer right click menu. This command will "refresh" the selected file(s) contents by exporting it from the currently connected server to the local file system. This is useful when one modifies the production and want to synchronise the changes from the server to local file system with ease.

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,7 +139,7 @@
         },
         {
           "command": "vscode-objectscript.refreshLocalFile",
-          "when": "editorLangId =~ /^objectscript/ && vscode-objectscript.connectActive"
+          "when": "resourceScheme == file && editorLangId =~ /^objectscript/ && vscode-objectscript.connectActive"
         },
         {
           "command": "vscode-objectscript.compileWithFlags",
@@ -475,11 +475,6 @@
           "group": "objectscript@3"
         },
         {
-          "command": "vscode-objectscript.refreshLocalFile",
-          "when": "editorLangId =~ /^objectscript/ && vscode-objectscript.connectActive",
-          "group": "objectscript@3"
-        },
-        {
           "command": "vscode-objectscript.previewXml",
           "when": "editorLangId =~ /^xml/",
           "group": "objectscript@4"
@@ -559,11 +554,6 @@
         },
         {
           "command": "vscode-objectscript.compileFolder",
-          "when": "vscode-objectscript.connectActive && resourceScheme == file",
-          "group": "objectscript_modify@1"
-        },
-        {
-          "command": "vscode-objectscript.refreshLocalFile",
           "when": "vscode-objectscript.connectActive && resourceScheme == file",
           "group": "objectscript_modify@1"
         },
@@ -807,11 +797,6 @@
       },
       {
         "category": "ObjectScript",
-        "command": "vscode-objectscript.refreshLocalFile",
-        "title": "Export class from server and replace current file"
-      },
-      {
-        "category": "ObjectScript",
         "command": "vscode-objectscript.previewXml",
         "title": "Preview XML as UDL"
       },
@@ -937,7 +922,7 @@
       {
         "category": "ObjectScript",
         "command": "vscode-objectscript.refreshLocalFile",
-        "title": "Refresh Contents"
+        "title": "Overwrite from Server"
       },
       {
         "category": "ObjectScript",
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "workspaceContains:**/*.csr",
     "onCommand:vscode-objectscript.export",
     "onCommand:vscode-objectscript.compile",
+    "onCommand:vscode-objectscript.refreshLocalFile",
     "onCommand:vscode-objectscript.viewOthers",
     "onCommand:vscode-objectscript.jumpToTagAndOffset",
     "onCommand:vscode-objectscript.previewXml",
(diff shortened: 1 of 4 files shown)


ollitanska:
> @ollitanska Thanks for the PR! Can you explain a little more why this PR is necessary? A user can re-export from the ObjectScript Explorer or using the export command and we are trying to avoid adding too many menu options.
> 
> cc @gjsjohnmurray

I understand that this feature will overlap with the current functionality, but I still believe that this PR would be beneficial to the users. Tho, totally understand if you decide not to accept this PR.

When working with local files, a another user can simultaneously modify the source code at the server and your local files can easily get out of sync. This feature would add a really handy and a quick way to re-sync the local file system state with the server. For my dev setup, out of sync will happen regularly, if I or another person edits the production settings in the web browser and forgets to export the changes to ones' local file system.

The ObjectScript Explorer re-export way have some drawbacks for this kind of development style / setup:
- If the user has only some of the files in their local file system (and doesn't want them all), export will add extra files that they have to delete after export
- If the user has the files located in different kind of structure that what the export expects (e.g. files spread around in multiple submodules and thus folders), the exported files are not placed in the correct folders. (Yes, I know that you can create a custom mapping for this, but my proposed way would be way simpler)

Perhaps, this functionality could be improved to not only re-export the server source files to the local file system, but to compare the timestamps between the two, and automatically import & export based on which one is the newer. Or maybe this could be achieved by having a toggle to enable automatic file sync between local file system and the server.

isc-bsaviano:
Thanks for taking the time to explain further. 

> When working with local files, another user can simultaneously modify the source code at the server and your local files can easily get out of sync.

Have you considered using a personal server for compilation and intellisense? This may not be possible in your case, but it's worth exploring since containerization and the community edition has made this much easier than in the past.

> The ObjectScript Explorer re-export way have some drawbacks for this kind of development style / setup:

These are two compelling reasons, with the caveat that web application files do need to be structured identically to the server for syncing them to work properly. I'm going to wait to see what John, one of our other maintainers, thinks before I dive into the code.

package.json
```diff
@@ -469,6 +474,11 @@
           "when": "editorLangId =~ /^objectscript/ && vscode-objectscript.connectActive",
           "group": "objectscript@3"
         },
+        {
+          "command": "vscode-objectscript.refreshLocalFile",
+          "when": "editorLangId =~ /^objectscript/ && vscode-objectscript.connectActive",
```
gjsjohnmurray, line 479:
```suggestion
          "when": "resourceScheme == file && editorLangId =~ /^objectscript/ && vscode-objectscript.connectActive",
```
Otherwise I think it would appear for isfs documents.

package.json
```diff
@@ -136,6 +137,10 @@
           "command": "vscode-objectscript.compile",
           "when": "editorLangId =~ /^objectscript/ && vscode-objectscript.connectActive"
         },
+        {
+          "command": "vscode-objectscript.refreshLocalFile",
+          "when": "editorLangId =~ /^objectscript/ && vscode-objectscript.connectActive"
```
gjsjohnmurray, line 142:
```suggestion
          "when": "resourceScheme == file && editorLangId =~ /^objectscript/ && vscode-objectscript.connectActive"
```
I don't think the command should be offered for an isfs document.

package.json
```diff
@@ -469,6 +474,11 @@
           "when": "editorLangId =~ /^objectscript/ && vscode-objectscript.connectActive",
           "group": "objectscript@3"
         },
+        {
```
gjsjohnmurray, line 477: My preference is for this command **not** to be added to the editor/context menu. That menu is already long, and users can't yet customize it. For people using a personal server namespace the command should be rarely needed. It's in Command Palette and a user can assign it a keyboard shortcut if they wish.

package.json
```diff
@@ -552,6 +562,11 @@
           "when": "vscode-objectscript.connectActive && resourceScheme == file",
           "group": "objectscript_modify@1"
         },
+        {
```
gjsjohnmurray, line 565: As commented re the editor/context menu, perhaps don't add this to explorer/context.

gjsjohnmurray (changes requested):
I haven't yet looked at the actual implementation of the command.

src/extension.ts
```diff
@@ -785,6 +791,24 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
     vscode.commands.registerCommand("vscode-objectscript.compileWithFlags", () => importAndCompile(true)),
     vscode.commands.registerCommand("vscode-objectscript.compileAll", () => namespaceCompile(false)),
     vscode.commands.registerCommand("vscode-objectscript.compileAllWithFlags", () => namespaceCompile(true)),
+    vscode.commands.registerCommand("vscode-objectscript.refreshLocalFile", async (_file, files) => {
```
isc-bsaviano, line 794: Since this command is only in the command palette now the implementation should be really simple, perhaps only a few lines:

```ts
const file = currentFile();
if (!file) {
  return;
}
loadChanges([file]);
```

IMO you also don't need to special Output channel messages anymore since you're only refreshing one file.

syntaxes/vscode-objectscript-output.tmLanguage.json
```diff
@@ -23,6 +23,14 @@
       "match": "^Compilation finished successfully in \\S+",
       "name": "markup.bold"
     },
+    {
```
isc-bsaviano, line 26: You can remove these changes now that this syntax is no longer used. Otherwise, looks good!

isc-bsaviano:
Thanks @ollitanska. Can you just run `npm run lint-fix` to resolve that code linting error? I will approve and merge after that.

isc-bsaviano (approved):

isc-bsaviano:
Thanks again for the PR!

PR #1080, opened 2023-02-07
Fix isfs folder deletion

### LINKED ISSUE #1079: Delete of subpackage over isfs fails
Connecting over isfs, if right click a subpackage by right clicking the folder for it in the workspace view, I get (for example, deleting SourceControl.Git, where multiple classes in that package exist):
`ERROR #16006: Document 'SourceControl.Git' name is invalid`
Oddly, deleting a top-level folder/package seems to work properly. It's just subpackages that don't work.

isc-bsaviano:
This PR fixes #1079

### DIFF
--- a/src/providers/FileSystemProvider/FileSystemProvider.ts
+++ b/src/providers/FileSystemProvider/FileSystemProvider.ts
@@ -442,7 +442,11 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
     if (fileName.startsWith(".")) {
       return;
     }
-    if (!fileName.includes(".")) {
+    if (
+      await this._lookupAsDirectory(uri)
+        .then(() => true)
+        .catch(() => false)
+    ) {
       // Get the list of documents to delete
       let toDeletePromise: Promise<any>;
       if (project) {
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (commented):
I'm slightly nervous about using _lookupAsDirectory() because if uri references a non-existent subdirectory I have a feeling the call will create an entry in a local structure which subsequent iteration of directories at that level will see as evidence the server actually has such a thing.

As long as delete() is being called for a uri that **does** exist on the server, no harm will result. And typically delete() only gets called when the user interacts with something they can see in Explorer, so it **will* be real on the server.

But if an extension called `vscode.workspace.fs.delete()`  with an isfs uri specifying a non-existent package I think the Explorer tree might subsequently show a corresponding folder.

I haven't verified my hunch, and I suppose the chances of it happening are slim.

isc-bsaviano:
@gjsjohnmurray Should I replace it with a call to `_lookup()` without the `fillInPath` parameter? That way a `FileNotFound` error will be thrown (which is the correct behavior) and if no errors are thrown I can check if the result is a `Directory`.

gjsjohnmurray (approved):
LGTM

PR #1088, opened 2023-02-10
Add 'Start debugging current method' option to launch configuration

ollitanska:
This PR adds a new variable/command to the objectscript launch configuration settings, which will allow the user to start debugging the current ClassMethod. I.e. allows user to press "Start debugging" / F5 to run the function of which the cursor in currently inside.

```
{
    "type": "objectscript",
    "request": "launch",
    "name": "ObjectScript: Debug current method",
    "program": "${command:CurrentMethod}"
},
````

### DIFF
--- a/package.json
+++ b/package.json
@@ -1522,7 +1522,8 @@
           }
         },
         "variables": {
-          "PickProcess": "vscode-objectscript.pickProcess"
+          "PickProcess": "vscode-objectscript.pickProcess",
+          "CurrentMethod": "vscode-objectscript.currentMethod"
         },
         "initialConfigurations": [
           {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -819,6 +819,69 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
           }
         });
     }),
+    vscode.commands.registerCommand("vscode-objectscript.currentMethod", async (config) => {
+      const file = currentFile();
+      if (!file.fileName.endsWith("cls")) {
+        vscode.window.showErrorMessage(
+          "Unable to start debugging as the cursor is not inside any debuggable method.",
+          "Dismiss"
+        );
+        return;
+      }
+
+      const symbols: vscode.DocumentSymbol[] = await vscode.commands.executeCommand<vscode.DocumentSymbol[]>(
+        "vscode.executeDocumentSymbolProvider",
+        file.uri
+      );
+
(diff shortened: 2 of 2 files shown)


isc-bsaviano:
We already have a `Debug` CodeLens for ClassMethods so I'm not sure this is necessary.

ollitanska:
> We already have a `Debug` CodeLens for ClassMethods so I'm not sure this is necessary.

Thank you for your comment. I understand that the CodeLens Debug feature already exists and I was actually inspired to create this change because I frequently use it.

One of the biggest advantages of this new feature is that it allows the user to 'trigger' the CodeLens Debug method with a shortcut, which provides greater control over how it is launched. With this, the user can now choose the CodeLens Debug as one of their debug targets, just like any other debug target. This preserves the go-to-way for configuring debug settings.

While it is possible to add a shortcut key for the CodeLens Debug, this would separate the functionality and configuration from the nominal way that VSCode does debugging. By incorporating this functionality, we can provide a more cohesive and integrated experience for the user.

isc-bsaviano:
@ollitanska U can see the value for users who love keybindings, but I don't think the way you implemented this is optimal. Your code requires that the user create a launch debug configuration, while the `Debug` CodeLens does that work for you. I think you should implement a new command that's only shown in the command palette for classes and routines that creates a debug session with `program` filled in using the current cursor's position. You could even add  a breakpoint at the cursor line and call it something like "Launch and Run to Current Line" or something. Here's how the CodeLens command is implemented:

https://github.com/intersystems-community/vscode-objectscript/blob/726be3b365dd3aa4ce6bbb4e0d9a040c95a565e9/src/extension.ts#L798-L821

ollitanska:
@isc-bsaviano Okay, sounds good. If the user wants, they can also add these commands to the launch debug configuration or add a custom keybinding for them I they want to. When I have time, I will do the required modifications.

ollitanska:
@isc-bsaviano I noticed that the VSCode already has a "Run to Cursor" command, which is equivalent to your "Launch and Run to Current Line" idea. Only problem with it is that the command relies on the launch debug configuration. But with my original PR, the user could use the already implemented "Run to Cursor" command if they so desire. I would feel a bit icky to implement an identical command that works only with ObjectScript.

image

isc-bsaviano:
Hi @ollitanska. I'm closing this PR since it's gotten stale. If you'd like to have it re-considered, please update your fork and respond to the comments above. Thanks again for your past pull requests!

PR #1090, opened 2023-02-13
Improve `isfs` folder creation/modification UX

isc-bsaviano:
IMO our existing command for creating an `isfs` workspace folder have 2 weaknesses:

- It can't be re-used to modify an existing folder. 
- It doesn't allow you to add query parameters. The Studio equivalents can be easily accessed in a menu.

This PR changes the command to allow for the addition of query parameters and for the most of the code to be re-used in a new "modify existing `isfs` folder" command. I also updated the docs to show the new UI.


gjsjohnmurray:
Based only on reading the diffs I found the new step shown in docs/assets/images/ss-query-params.png hard to understand. Seems like I either set only the first checkbox, or only the second, or neither of those but zero or more of the others.

Did you rule out having a single-choice quickpick to handle ordinary, csp or project, followed by the multi-select one if the choice was o"ordinary"?

Did you consider making it a multi-step procedure, so I can back up to a previous step?

isc-bsaviano:
@gjsjohnmurray Looking at this with fresh eyes I see the confusion. I think splitting it up into 2 steps makes sense, but I'm not sure what the call the "ordinary" step that leads to the other query params, or what the prompt for that first step should be.

isc-rsingh:
Can't look at this until tomorrow AM

isc-rsingh:
Since project rules out other choices, that should be the first step (title="Open Project", options=yes/no). Then if project isn't selected, go to CSP (title="Show web application files", options=yes/no), then type or filter.

isc-bsaviano:
I prefer John's 2 step approach if I can get the labels right. I'll keep working on this and update the PR when I think of something.

isc-bsaviano:
@gjsjohnmurray @isc-rsingh I just pushed a new commit that split up the query parameter QuickPick and changed the language a little. Let me know what you think.

isc-rsingh:
Much improved

isc-rsingh (approved):

isc-bsaviano:
@gjsjohnmurray If you don't have any comments I'm going to merge this.

gjsjohnmurray:
@isc-bsaviano I'm busy prepping for a webinar we're giving later today. Should be able to look at PRs tomorrow.

src/commands/addServerNamespaceToWorkspace.ts
```diff
@@ -179,3 +147,191 @@ export async function getServerManagerApi(): Promise<any> {
...
+  // Prompt the user for the files to show
+  const filterType = await vscode.window.showQuickPick(
+    [
+      {
+        label: "$(file-code) Show web application files.",
+        detail: "Choose a specific web application, or show all.",
+        value: "csp",
+      },
+      {
+        label: "$(files) Show the contents of a server-side project.",
```
gjsjohnmurray, line 167:
```suggestion
        label: "$(files) Contents of a Server-side Project",
```

src/commands/addServerNamespaceToWorkspace.ts
```diff
@@ -179,3 +147,191 @@ export async function getServerManagerApi(): Promise<any> {
...
+        detail: "Choose an existing project or create a new one.",
+        value: "project",
+      },
+      {
+        label: "$(list-tree) Show files that pass through custom filters.",
```
gjsjohnmurray, line 172:
```suggestion
        label: `$(list-tree) Code Files in ${namespace}`,
```

src/commands/addServerNamespaceToWorkspace.ts
```diff
@@ -179,3 +147,191 @@ export async function getServerManagerApi(): Promise<any> {
...
+  const filterType = await vscode.window.showQuickPick(
+    [
+      {
+        label: "$(file-code) Show web application files.",
+        detail: "Choose a specific web application, or show all.",
+        value: "csp",
+      },
+      {
+        label: "$(files) Show the contents of a server-side project.",
+        detail: "Choose an existing project or create a new one.",
```
gjsjohnmurray, line 168:
```suggestion
        detail: "Choose an existing project, or create a new one.",
```

src/commands/addServerNamespaceToWorkspace.ts
```diff
@@ -179,3 +147,191 @@ export async function getServerManagerApi(): Promise<any> {
...
+  if (filterType.value == "csp") {
+    // Prompt for a specific web app
+    const cspApps = cspAppsForUri(uri);
+    if (cspApps.length == 0) {
+      vscode.window.showInformationMessage("No web applications are configured in this namespace.", "Dismiss");
+      return;
+    }
+    newPath =
+      (await vscode.window.showQuickPick(cspApps, {
+        title: "Pick a specific web application to show, or press 'Escape' to show all.",
```
gjsjohnmurray, line 197:
```suggestion
        placeHolder: "Pick a specific web application to show, or press 'Escape' to show all",
```

src/commands/addServerNamespaceToWorkspace.ts
```diff
@@ -179,3 +147,191 @@ export async function getServerManagerApi(): Promise<any> {
...
+        value: "csp",
+      },
+      {
+        label: "$(files) Show the contents of a server-side project.",
+        detail: "Choose an existing project or create a new one.",
+        value: "project",
+      },
+      {
+        label: "$(list-tree) Show files that pass through custom filters.",
+        detail: "Choose the filters to apply.",
```
gjsjohnmurray, line 173:
```suggestion
        detail: "Filters can be applied in the next step.",
```

src/commands/addServerNamespaceToWorkspace.ts
```diff
@@ -179,3 +147,191 @@ export async function getServerManagerApi(): Promise<any> {
...
+    return;
+  }
+
+  let newParams = "";
+  let newPath = "/";
+  if (filterType.value == "csp") {
+    // Prompt for a specific web app
+    const cspApps = cspAppsForUri(uri);
+    if (cspApps.length == 0) {
+      vscode.window.showInformationMessage("No web applications are configured in this namespace.", "Dismiss");
```
gjsjohnmurray, line 192:
```suggestion
      vscode.window.showWarningMessage(`No web applications are configured to use namespace ${namespace}.`, "Dismiss");
```

src/commands/addServerNamespaceToWorkspace.ts
```diff
@@ -179,3 +147,191 @@ export async function getServerManagerApi(): Promise<any> {
...
+        {
+          label: "$(references) mapped",
+          detail: "Hide files that are mapped from a non-default database.",
+          picked: oldParams.has("mapped"),
+        },
+      ],
+      {
+        ignoreFocusOut: true,
+        canPickMany: true,
+        title: "",
```
gjsjohnmurray, line 236:
```suggestion
        placeHolder: "Add optional filters",
```

src/commands/addServerNamespaceToWorkspace.ts
```diff
@@ -179,3 +147,191 @@ export async function getServerManagerApi(): Promise<any> {
...
+    const otherParams = await vscode.window.showQuickPick(
+      [
+        {
+          label: "$(filter) filter",
+          detail: "Comma-delimited list of search options. (i.e. '*.cls,*.inc,*.mac,*.int')",
+          picked: oldParams.has("filter"),
+        },
+        {
+          label: "$(list-flat) flat",
+          detail: "Show a flat list of files. Does not split packages as folders.",
```
gjsjohnmurray, line 219:
```suggestion
          detail: "Show a flat list of files. Do not treat packages as folders.",
```

src/commands/addServerNamespaceToWorkspace.ts
```diff
@@ -179,3 +147,191 @@ export async function getServerManagerApi(): Promise<any> {
...
+      return;
+    }
+    newParams = `project=${project}`;
+  } else {
+    // Prompt the user for other query parameters
+    const otherParams = await vscode.window.showQuickPick(
+      [
+        {
+          label: "$(filter) filter",
+          detail: "Comma-delimited list of search options. (i.e. '*.cls,*.inc,*.mac,*.int')",
```
gjsjohnmurray, line 214:
```suggestion
          detail: "Comma-delimited list of search options, e.g. '*.cls,*.inc,*.mac,*.int'",
```

src/commands/addServerNamespaceToWorkspace.ts
```diff
@@ -179,3 +147,191 @@ export async function getServerManagerApi(): Promise<any> {
...
+          picked: oldParams.has("flat"),
+        },
+        {
+          label: "$(server-process) generated",
+          detail: "Show generated files as well as non-generated.",
+          picked: oldParams.has("generated"),
+        },
+        {
+          label: "$(references) mapped",
+          detail: "Hide files that are mapped from a non-default database.",
```
gjsjohnmurray, line 229:
```suggestion
          detail: `Hide files that are mapped into ${namespace} from another code database.`,
```

src/commands/addServerNamespaceToWorkspace.ts
```diff
@@ -179,3 +147,191 @@ export async function getServerManagerApi(): Promise<any> {
...
+          picked: oldParams.has("filter"),
+        },
+        {
+          label: "$(list-flat) flat",
+          detail: "Show a flat list of files. Does not split packages as folders.",
+          picked: oldParams.has("flat"),
+        },
+        {
+          label: "$(server-process) generated",
+          detail: "Show generated files as well as non-generated.",
```
gjsjohnmurray, line 224:
```suggestion
          detail: "Also show files tagged as generated, e.g. by compilation.",
```

src/commands/addServerNamespaceToWorkspace.ts
```diff
@@ -179,3 +147,191 @@ export async function getServerManagerApi(): Promise<any> {
...
+  if (!wsFolderUri) {
+    // Select a workspace folder to modify
+    if (vscode.workspace.workspaceFolders == undefined || vscode.workspace.workspaceFolders.length == 0) {
+      vscode.window.showErrorMessage("No workspace folders are open.", "Dismiss");
+      return;
+    } else if (vscode.workspace.workspaceFolders.length == 1) {
+      wsFolder = vscode.workspace.workspaceFolders[0];
+    } else {
+      wsFolder = await vscode.window.showWorkspaceFolderPick({
+        placeHolder: "Pick the workspace folder modify.",
```
gjsjohnmurray, line 287:
```suggestion
        placeHolder: "Pick the workspace folder to modify",
```

gjsjohnmurray (changes requested):
I have made some inline suggestions.

I also think this could be improved:

image

The short lowercase labels feel a bit cryptic. I would prefer:
- Filter
- Flat List
- Generated Code
- Mapped Code

Either your switch statement would need to be amended as well, or you could extend QuickPickItem to add, say, a `value` property to each option and use that in your switch.

I also find it confusing that on this dialog `generated` is about **showing** and `mapped` is about **hiding**. Under the covers both are about showing. How about pre-selecting mapped if it's a new definition or if the existing definition doesn't specify a value? Alternatively make the option labels be "Show Generated" and "Hide Mapped".

isc-bsaviano:
@gjsjohnmurray  `mapped` is hide but `generated` is show because by default StudioOpenDialog shows `mapped` files but hides `generated` ones. I can either switch the names like you suggested, or I can select mapped by default and handle that item separately.

src/commands/addServerNamespaceToWorkspace.ts
```diff
@@ -80,82 +81,49 @@ export async function addServerNamespaceToWorkspace(): Promise<void> {
...
+        label: `$(pencil) Edit Code in ${namespace}`,
+        detail: "Documents opened in this folder will be editable.",
+      },
+      {
+        value: FILESYSTEM_READONLY_SCHEMA,
+        label: `$(lock) View Code in ${namespace}`,
+        detail: "Documents opened in this folder will be read-only.",
+      },
+    ],
+    { title: "Choose the type of access", ignoreFocusOut: true }
```
gjsjohnmurray, line 98:
```suggestion
    { placeHolder: "Choose the type of access", ignoreFocusOut: true }
```
For consistency with the other steps.

src/commands/addServerNamespaceToWorkspace.ts
```diff
@@ -179,3 +147,192 @@ export async function getServerManagerApi(): Promise<any> {
...
+      switch (otherParamName) {
+        case "filter": {
+          // Prompt for filter
+          const filter = await vscode.window.showInputBox({
+            title: "Enter a filter string.",
+            ignoreFocusOut: true,
+            value: oldParams.get("filter"),
+            placeHolder: "*.cls,*.inc,*.mac,*.int",
+            prompt:
+              "Patterns are comma-delimited and may contain both * (any number of characters) and ? (a single character) as wildcards. To exclude items, prefix the pattern with a single quote.",
```
gjsjohnmurray, line 256:
```suggestion
              "Patterns are comma-delimited and may contain both * (zero or more characters) and ? (a single character) as wildcards. To exclude items, prefix the pattern with a single quote.",
```

src/commands/addServerNamespaceToWorkspace.ts
```diff
@@ -179,3 +147,192 @@ export async function getServerManagerApi(): Promise<any> {
...
+  }
+
+  // Prompt the user to modify the uri
+  const newUri = await modifyWsFolderUri(wsFolder.uri);
+  if (!newUri) {
+    return;
+  }
+  // Prompt for name change
+  const newName = await vscode.window.showInputBox({
+    title: "Enter a name for the workspace folder.",
```
gjsjohnmurray, line 317:
```suggestion
    title: "Enter a name for the workspace folder",
```

gjsjohnmurray (changes requested):
A couple more tweaks.

Re 'mapped', I don't mind whether you opt to change the label or pre-select the checkbox.

Also, when testing the modify path I noticed that the "what to show" quickpick defaults to the first item even though the URI I'm modifying was created using the third item (i.e. it's a regular one for code in a namespace). I think you could pre-select the right item, but only if you use `createQuickPick` and friends rather than the convenience of `showQuickPick`.

While on the subject of that particular quickpick, how about changing the order so "Code Files in XXX" comes first?

gjsjohnmurray:
I just discovered that if I'm trying to add a web-app-type folder from a server I haven't already connected to, I get told that there are no web apps on the server which use the namespace I chose. This happens because `cspAppsForUri` relies on information in the `cspApps` map which only gets put there by an `await checkConnection(false, uri)` which hasn't yet happened.

isc-bsaviano:
@gjsjohnmurray Good catch, I will change that to fetch the list from the server

gjsjohnmurray:
When I use the Modify feature on a uri that I had manually added `?system=1` this gets dropped.

Did you consider adding a filter option for including system items? Either way, existing queryparams that the dialog doesn't handle should be preserved IMO.

Also noticed that if I start a modify but don't change anything on the way through I get told "Failed to modify...".

And I suggest we be consistent about capitalization of `Server-Side` (in the Modify command vs `Server-side` (in the quickpick option about projects).

isc-bsaviano:
@gjsjohnmurray I didn't create an entry for `system` because it's undocumented. I can do that though. The only other param is `type`, which I do not want to create an entry for.

gjsjohnmurray:
Latest build no longer seems to update the URI when I use the Modify command. Here's my workspace definition:

```json
{
	"folders": [
		{
			"name": "iris231:USER",
			"uri": "isfs://iris231:user/?mapped=0&system=1"
		}
	],
	"settings": {
		"objectscript.showExplorer": false
	}
}
```

When I get to the checkboxes in the Modify pathway the last two are correctly checked. But if I uncheck either or both of these, OK the quickpick, OK the name (changed or unchanged) the uri in my workspace definition doesn't change.

isc-bsaviano:
@gjsjohnmurray Sorry about that, my last commit should fix it. I now clear out all known query parameters but leave intact ones that I don't know (like `type`).

isc-bsaviano:
Sorry, that got lost in the shuffle. I'll take a look

isc-bsaviano:
@gjsjohnmurray Done

gjsjohnmurray (approved):
LGTM

gjsjohnmurray:
Too late for this PR I know, but maybe this step would be clearer with an explicit "<all applications>" entry in the quickpick:

image

isc-bsaviano:
That's a good suggestion! Maybe an entry called `All`?

PR #1092, opened 2023-02-14
Add charset to REST Debug panel's Content-Type

### LINKED ISSUE #1091: HTTP headers don't work correctly on the Debug REST Service if it contains uppercase letters.
When I set "Content-Type:application/json;charset=utf-8" in the HEADER, non-ascii characters in the BODY are garbled in the VALIABLES window.
image

image

But I set all lowercase "content-type:application/json;charset=utf-8", I got correct view as follows:
image
image

[Http headers are case insensitive.](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers)

It's good for developers if we can write case insensitive in the HEADER of the Debug REST Service.

isc-bsaviano:
This PR fixes #1091 

Generally, only `text/*` MIME types can contain the `charset` parameter but IRIS seems to respond to it for JSON so I added it for all types.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#structure_of_a_mime_type

### DIFF
--- a/src/commands/restDebugPanel.ts
+++ b/src/commands/restDebugPanel.ts
@@ -391,16 +391,18 @@ export class RESTDebugPanel {
             message.headersText.split(/\r?\n/).forEach((line) => {
               line = line.trim();
               if (line != "" && !line.startsWith("#")) {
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (approved):
LGTM but I haven't tested it.

PR #1097, opened 2023-02-21
Fix namespace pick when trying to connect without permissions on `%SYS`

### LINKED ISSUE #1096: No connection with roleset that can't see %SYS
When trying to connect to a server with a user that does not have the privileges required to access, or even "see", the %SYS namespace, connection fails **immediately**, with an **instant** error message 

`This server does not have specified namespace '%SYS'. You must select one of the following: USER.`

Usually, with user that are able to see and use %SYS, we get a list of namespaces available to connect to, with the problematic case it goes straight to the error message and does not provide a dropdown list. %SYS was never specified in server settings or during attempts to connect.

isc-bsaviano:
This PR fixes #1096

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -439,12 +439,12 @@ export class AtelierAPI {
     }
   }
 
-  public serverInfo(): Promise<Atelier.Response<Atelier.Content<Atelier.ServerInfo>>> {
+  public serverInfo(checkNs = true): Promise<Atelier.Response<Atelier.Content<Atelier.ServerInfo>>> {
     return this.request(0, "GET").then((info) => {
       if (info && info.result && info.result.content && info.result.content.api > 0) {
         const data = info.result.content;
         const apiVersion = data.api;
-        if (this.ns && this.ns.length && !data.namespaces.includes(this.ns)) {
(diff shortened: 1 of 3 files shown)


src/api/index.ts
```diff
@@ -439,12 +439,12 @@ export class AtelierAPI {
     }
   }
 
-  public serverInfo(): Promise<Atelier.Response<Atelier.Content<Atelier.ServerInfo>>> {
+  public serverInfo(checkNs = true): Promise<Atelier.Response<Atelier.Content<Atelier.ServerInfo>>> {
```
gjsjohnmurray, line 442:
```suggestion
  public serverInfo(checkNs: boolean = true): Promise<Atelier.Response<Atelier.Content<Atelier.ServerInfo>>> {
```

gjsjohnmurray (changes requested):
Just one minor suggestion added.

src/api/index.ts
isc-bsaviano: ESLint strips that type annotation out when you run `mpn run lint-fix`.

gjsjohnmurray: Fair enough.

gjsjohnmurray (approved):

PR #1105, opened 2023-02-24
Prepare next release

isc-bsaviano:
The next version is currently set for 2.4.4 but the change list is pretty big so if you want to bump it to 2.6.0 I can do that.


gjsjohnmurray:
I think this deserves to be 2.6.0.

I also suggest that the item about async server search mentions this only happens on 2023.1+, and that all server search still requires the procedure to use proposed API.

gjsjohnmurray (approved):

PR #1107, opened 2023-03-03
Add `SOAP Wizard` to Server Actions menu

isc-bsaviano:
This PR adds the `SOAP Wizard` to the Server Actions menu. We get a lot of questions about this wizard so it makes sense that we should link to it without requiring user configuration. If the user already manually added a link to it using the `objectscript.conn.links` setting, their menu item will be shown instead of our default one.


gjsjohnmurray (approved):
LGTM

PR #1116, opened 2023-03-13
Deprecate confusing `objectscript.serverSideEditing` setting

### LINKED ISSUE #1115: Consider removing `objectscript.serverSideEditing` setting
I think we should remove the `objectscript.serverSideEditing` configuration setting for the following reasons:

- Users will confuse it with [server-side editing mode](https://intersystems-community.github.io/vscode-objectscript/serverside/), leading to them using a sub-optimal workflow.
- The setting blurs the line between client-side and server-side editing. Having two supported editing paradigms is a major feature of this extension, but having this middle-ground setting makes documenting more difficult. It doesn't make sense conceptually to have the ObjectScript Explorer open editable server-side files when the workflow dictates that you should be exporting files to edit.
- Users who enable this setting may make edits that are not tracked by their client-side source control, leading to pain down the road as things don't work the same way on different systems. 

The only downside to removing this setting is that some existing users are using it, so their workflows may be disrupted. However, I think the benefits outweigh that temporary friction and that those users will be happier being firmly in one of the paradigms. 

@gjsjohnmurray @isc-rsingh @daimor What do you think?

isc-bsaviano:
This PR fixes #1115

### DIFF
--- a/docs/ServerSide.md
+++ b/docs/ServerSide.md
@@ -178,7 +178,6 @@ Use **File > New File** to create a new file. Add content similar to the followi
     }
   ],
   "settings": {
-    "objectscript.serverSideEditing": true
   }
 }
 ```
@@ -220,8 +219,7 @@ Example with connection to different namespaces on the same server.
       "password": "SYS",
       "ns": "MYAPP",
       "port": 52773,
-    },
-    "objectscript.serverSideEditing": true
+    }
   }
 }
 ```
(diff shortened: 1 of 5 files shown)


isc-rsingh (approved):

gjsjohnmurray:
I wonder if it might be wiser to do this in two stages, first removing the setting from documentation and package.json and displaying a notification to any user who has set it true, saying it will be removed in the next release.

isc-bsaviano:
Sure, I can do that

isc-rsingh (approved):

PR #1124, opened 2023-03-22
For testingmanager

gjsjohnmurray:
This PR implements changes required by the upcoming testing manager extension.

### DIFF
--- a/src/commands/compile.ts
+++ b/src/commands/compile.ts
@@ -96,6 +96,11 @@ async function importFile(
   if (typeof file.content === "string") {
     enc = false;
     content = file.content.split(/\r?\n/);
+
+    // Avoid appending a blank line on every save, which would cause a web app file to grow each time
+    if (content.length > 1 && content[content.length - 1] === "") {
+      content.pop();
+    }
   } else {
     // Base64 encoding must be in chunk size multiple of 3 and within the server's potential 32K string limit
     // Output is 4 chars for each 3 input, so 24573/3*4 = 32764
--- a/src/commands/compile.ts
+++ b/src/commands/compile.ts
@@ -351,6 +351,10 @@ export async function importAndCompile(
             fileSystemProvider.fireFileChanged(file.uri);
           }
         }
+      } else if (file.uri.scheme === FILESYSTEM_SCHEMA || file.uri.scheme === FILESYSTEM_READONLY_SCHEMA) {
+        // Fire the file changed event to avoid VSCode alerting the user on the next folder-specific save (e.g. of settings.json) that
+        // "The content of the file is newer."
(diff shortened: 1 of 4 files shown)


src/providers/FileSystemProvider/FileSystemProvider.ts
```diff
@@ -24,7 +24,8 @@ export type Entry = File | Directory;
 export function generateFileContent(fileName: string, sourceContent: Buffer): { content: string[]; enc: boolean } {
   const sourceLines = sourceContent.toString().split("\n");
   const fileExt = fileName.split(".").pop().toLowerCase();
-  if (fileExt === "cls") {
+  const csp = fileName.startsWith("/");
```
isc-bsaviano, line 27: Why are these modifications necessary? Can classes and routines be stored in a web application folder? That seems weird and a bad idea.

src/commands/compile.ts
```diff
@@ -346,6 +351,10 @@ export async function importAndCompile(
             fileSystemProvider.fireFileChanged(file.uri);
           }
         }
+      } else if (file.uri.scheme === FILESYSTEM_SCHEMA || file.uri.scheme === FILESYSTEM_READONLY_SCHEMA) {
+        // Fire the file changed event to avoid VSCode alerting the user on the next folder-specific save (e.g. of settings.json) that
+        // "The content of the file is newer."
+        fileSystemProvider.fireFileChanged(file.unredirectedUri ?? file.uri);
```
isc-bsaviano, line 357: Is this specifically for the VS Code json files? I don't think Ive seen this problem otherwise.

src/providers/FileSystemProvider/FileSystemProvider.ts
```diff
@@ -554,7 +557,7 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
     }
     // Write the new file
     // This is going to attempt the write regardless of the user's response to the check out prompt
-    const api = new AtelierAPI(oldUri);
+    const api = new AtelierAPI(newUri);
```
isc-bsaviano, line 560: What bug is this fixing? We prevent renaming across workspace folders so the connection info should be the same between `oldUri` and `newUri`.

src/commands/compile.ts
gjsjohnmurray: It showed up when using vscode.fs.copy() to put a testclass file into the server-side folder %UnitTest.Manager runs it from.

src/providers/FileSystemProvider/FileSystemProvider.ts
gjsjohnmurray: When Testing Manager copies a .cls file from a file:// uri to an isfs:// one.

PR #1156, opened 2023-05-26
Make WebSocket Terminal launchable from Command Palette and Server Manager tree

gjsjohnmurray:
This PR fixes a bug with the command for launching a terminal. It also contributes launch buttons to namespaces in the Server Manager tree, and respects the `terminal.integrated.defaultLocation` setting.


isc-bsaviano (changes requested):

isc-bsaviano (approved):
Haven't tested but code looks good now

daimor (approved):

PR #1208, opened 2023-08-10
Remove `flat` and `type` isfs query parameters

isc-bsaviano:
This PR removes the query parameters that were deprecated in #1165


gjsjohnmurray (approved):
LGTM

PR #1211, opened 2023-08-17
Improve message to help resolve scenario where isfs user lacks %DB_IRISSYS:READ

gjsjohnmurray:
If an isfs user only holds the out-of-the-box %Developer role they aren't able to access the %SYS namespace, which is the home of the webapp providing folder-specific settings storage. This scenario was producing a misleading message advising a SQL GRANT statement to be run. This PR improves the message to help identify the problem.

### DIFF
--- a/src/providers/FileSystemProvider/FileSystemProvider.ts
+++ b/src/providers/FileSystemProvider/FileSystemProvider.ts
@@ -259,9 +259,10 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
         if (error) {
           console.log(error);
           if (error.errorText.includes(" #5540:")) {
+            const nsUpper = api.config.ns.toUpperCase();
             const message = `User '${api.config.username}' cannot list ${
-              csp ? "web application" : "namespace"
-            } contents. To resolve this, execute the following SQL in the ${api.config.ns.toUpperCase()} namespace:\n\t GRANT EXECUTE ON %Library.RoutineMgr_StudioOpenDialog TO ${
+              csp ? `web application '${uri.path}'` : "namespace"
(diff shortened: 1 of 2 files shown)


src/providers/FileSystemProvider/FileSystemProvider.ts
```diff
@@ -259,9 +259,10 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
         if (error) {
           console.log(error);
           if (error.errorText.includes(" #5540:")) {
+            const nsUpper = api.config.ns.toUpperCase();
             const message = `User '${api.config.username}' cannot list ${
-              csp ? "web application" : "namespace"
-            } contents. To resolve this, execute the following SQL in the ${api.config.ns.toUpperCase()} namespace:\n\t GRANT EXECUTE ON %Library.RoutineMgr_StudioOpenDialog TO ${
+              csp ? `web application '${uri.path}'` : "namespace"
```
isc-bsaviano, line 264: This could look odd if uri.path is "" (the root).

gjsjohnmurray: I think for the root, uri.path="/"

In any case, I think this situation is very unlikely to occur. ISFS workspaces are typically created from the Server Manager tree, and a user without read permission on a namespace's code database wouldn't get to see the namespace in that tree, nor therefore the web apps (possibly "/") associated with it. The "/_vscode/XXX" path that's used to seek the .vscode/settings.json file for namespace XXX is atypical because the user doesn't browse the SM tree for it, and it comes from the "/_vscode" web app which, if defined, belongs to the %SYS namespace.

isc-bsaviano (approved):

PR #1232, opened 2023-09-15
Make it easier to add namespace from same server to workspace

gjsjohnmurray:
When using the Explorer context menu on a folder in an isfs workspace to add another root, make it easier to pick another namespace on the same server as the current root.

image


gjsjohnmurray:
> It doesn't reduce prompts for the case where you want to use the same server

It almost always reduces keystrokes for that case, the exception being when the desired server is first in your list, in which case you use the same number of keystrokes as before.

Yes, long server lists can be filtered but it seems unhelpful not to have a quick path for the "another namespace on this same server" case.

Yes, it adds 2 keystrokes to the other path.

isc-bsaviano (approved):
Your comment makes sense.

PR #1236, opened 2023-09-22
Add option in Server Actions menu to open Studio Add-Ins

isc-bsaviano:
This PR adds an option to the Server Actions menu for selecting a Studio Add-In to open. It queries the server for the list of all Add-Ins, so it should support any customer-created Add-Ins (if that's even a thing). I removed the hard-coded SOAP Wizard option added by #1107 since that functionality is covered by the new option.


gjsjohnmurray (approved):
Nice!

PR #1253, opened 2023-10-17
Don't overwrite file name if the new name couldn't be determined

### LINKED ISSUE #1233: Why does this plugin automatically alter the contents of my cls files?
Whenever I drag a cls file from the File Explorer into the Project Folder in VSCode, or drag a cls file from one project folder to another, it automatically changes the content of my files, attempting to replace the package and class name, even if my directory isn't a Server directory. I haven't found any setting to control this behavior.

Specifically, some other software, like JReport, also use cls as an extension. This creates an issue when I try to drag such files into the project directory; they get overwritten, forcing me to manually copy them using the File Explorer.

I understand the initial intent behind this feature is to manage namespaces through the local folder structure, similar to many Java IDEs. However, even those IDEs display a warning or prompt instead of silently altering user files when they are located in an incorrect directory.

Also, during my work, I often need to modify only a part of the same file across multiple environments (while keeping other parts different). I frequently keep multiple versions of the same file in the same project directory, as illustrated below. In such cases, this drag-and-drop file renaming becomes quite inconvenient.

image

So, I'd appreciate it if you could add a toggle for this feature. That way, I could turn it off if I don't require it. 
(Furthermore, if possible, consider implementing a select option that allows different levels of behavior, including auto-modification, red highlighting, issuing a warning, or completely ignoring it.)

isc-bsaviano:
This PR fixes #1233

### DIFF
--- a/src/providers/FileSystemProvider/FileSystemProvider.ts
+++ b/src/providers/FileSystemProvider/FileSystemProvider.ts
@@ -32,49 +32,66 @@ export function generateFileContent(
   const csp = fileName.startsWith("/");
   if (fileExt === "cls" && !csp) {
     const className = fileName.split(".").slice(0, -1).join(".");
-    const content: string[] = [];
+    let content: string[] = [];
     const preamble: string[] = [];
 
-    // If content was provided (e.g. when copying a file), use all lines except for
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (approved):
LGTM but I didn't test it

PR #1255, opened 2023-10-19
Adjust activity bar icon to work correctly with upcoming 1.84 top-bar feature

gjsjohnmurray:
Already in 1.84 Insiders the Activity Bar position can be changed from Side to Top. The SVG of the InterSystems logo that the ObjectScript Explorer contributes wasn't displaying properly because its ratio wasn't 1:1

This PR corrects that.

Before:

image

After:

image

### DIFF
--- a/images/InterSystems.svg
+++ b/images/InterSystems.svg
@@ -1,25 +1,25 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <svg
-   class="custom-logo "
-   version="1.1"
-   id="Layer_1"
-   width="128px"
-   height="128px"
-   viewBox="0 0 128 128"
-   xml:space="preserve"
-   xmlns="http://www.w3.org/2000/svg"
-   xmlns:svg="http://www.w3.org/2000/svg">
+  class="custom-logo "
+  version="1.1"
+  id="Layer_1"
+  x="0px"
+  y="0px"
+  viewBox="0 0 80 80"
+  xml:space="preserve"
+  xmlns="http://www.w3.org/2000/svg"
+  xmlns:svg="http://www.w3.org/2000/svg">
   <g
    id="Canvas"
    fill="none"
-   transform="matrix(1.6318908,0,0,1.1754735,-22.624497,-0.68381501)">
-   	  <polygon
-   fill="#ffffff"
-   points="32.6,9.1 32.6,95.7 60.2,109.4 60.2,94.1 46.4,87.2 46.4,15.8 "
-   id="polygon1" />
-   	  <polygon
-   fill="#ffffff"
-   points="73.8,101 73.8,14.4 46.4,0.6 46.4,15.8 60.2,22.9 60.2,94.1 "
-   id="polygon2" />
-    </g>
-  </svg>
+   transform="matrix(1.1764706,0,0,1.1764706,15.338235,-9.2941177)">
+    <polygon
+      fill="#ffffff"
+      points="25.1,75.9 25.1,66.3 16.5,62 16.5,17.4 7.9,13.2 7.9,67.3 "
+      id="polygon1" />
+    <polygon
+    fill="#ffffff"
+    points="16.5,7.9 16.5,17.4 25.1,21.8 25.1,66.3 33.6,70.6 33.6,16.5 "
+    id="polygon2" />
+  </g>
+</svg>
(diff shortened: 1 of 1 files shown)


gjsjohnmurray:
@isc-bsaviano @isc-rsingh what do you think of this change? And is it time for a new release?

isc-rsingh (dismissed):

isc-bsaviano:
@gjsjohnmurray Sorry John, I have been on vacation. I tested this quickly and I don't like the change. The icon is too small on the regular activity bar.

isc-rsingh:
Now that I take the time to actually run this version I agree w @isc-bsaviano. Let me see if I can get it bigger.

gjsjohnmurray:
Upsized to 128x128 and scaled logo to use full height.

image

image

isc-bsaviano:
@gjsjohnmurray The logo still looks off to me.

@isc-rsingh I think you should reahc out to marketing or legal and se if they can give us an SVG that meets [these requirements](https://code.visualstudio.com/api/references/contribution-points#Icon-specifications).

isc-rsingh:
I spent a couple hours on this yesterday. I created a 24x24 SVG icon (I assume the docs mean pixels although the SVG for GitHub pull requests extension specifies pt not px) but it still appeared too small. I'm going to check out a few more other extensions to see what they do...

gjsjohnmurray:
@isc-bsaviano you maybe have an unrealistic expectation of what is possible because of how the existing one breaks the 1:1 ratio rule. By being twice as high as it is wide it seems to get scaled based on its width, then overflows above and below the area VS Code expects it to occupy.

I suggest one of us creates a solid 24x24 SVG (or it could be, say, 128x128) to illustrate the largest the ISC icon can be.

isc-rsingh:
> I suggest one of us creates a solid 24x24 SVG (or it could be, say, 128x128) to illustrate the largest the ISC icon can be.

@gjsjohnmurray here that is. I'm going to try to understand how others are getting bigger icons displayed but I'm swamped until tomorrow.

<?xml version="1.0" encoding="UTF-8"?>
<svg width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <g id="Canvas">
        <path id="Path" fill="#ffffff" stroke="none" d="M 10.123636 3.447273 L 7.112727 1.985455 L 7.112727 20.879999 L 13.134545 23.869091 L 13.134545 20.530909 L 10.123636 19.025454 Z"/>
        <path id="path1" fill="#ffffff" stroke="none" d="M 13.134545 20.530909 L 16.101818 22.036364 L 16.101818 3.141819 L 10.123636 0.130909 L 10.123636 3.447273 L 13.134545 4.996365 Z"/>
    </g>
</svg>

gjsjohnmurray:
Thanks @isc-rsingh. My hunch is that the ones using codicons get a slightly larger icon than the ones using SVGs.

gjsjohnmurray:
This is what I meant. The three highlighted are SVGs. The others are codicons.

image

image

But it turns out both the codicons and the SVG ones are displaying the same size. Here's the second screenshot with 30x30 squares overlaid (logically 24x24, but screenshot is from a 2560x1440 Windows display set to scale to 125%):

image


This is what InterSystems.svg contains for the above screenshots:
```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg
   width="24"
   height="24"
   viewBox="0 0 24 24"
   version="1.1"
   id="svg1"
   xmlns="http://www.w3.org/2000/svg"
   xmlns:svg="http://www.w3.org/2000/svg">
  <rect
     style="fill:#ffffff"
     id="rect1"
     width="24"
     height="24"
     x="0"
     y="0" />
</svg>
```

gjsjohnmurray:
With the current published extension (2.10.3) here's a 400% zoom in Paint of a section of the Activity Bar on the same 125% scaled display, with 30x30 outlines added.

image

isc-bsaviano:
Thanks for those images, John. So the current icon is too tall, but it doesn't get clipped in the activity bar due to a happy accident. The icon in https://github.com/intersystems-community/vscode-objectscript/pull/1255#issuecomment-1775886581 is the correct height, but not scaled properly so it appears too wide. Could we create an icon with 30 pixel height but the same scale as the current one? Sorry, I'm an SVG novice.

gjsjohnmurray:
How about this?

image

image

And this second one in Paint at 400% with 30x30 (i.e. 24x24) outlines:

image

gjsjohnmurray:
And the same on an unscaled display (i.e. true 24x24px)

image

image

isc-bsaviano (approved):
Thanks for the images John. It's a shame that we can't use the "big" logo for the activity bar and the "small" one for non-standard places. I'm going to approve this because I think it's the best we can do, but I think we should wait for Raj to approve as well before merging.

gjsjohnmurray:
OK, thanks @isc-bsaviano. I will wait for a review from @isc-rsingh

isc-rsingh:
I looked at a couple more extensions and I agree this is the best we can do. Our logo looks smaller because it's so narrow and there's nothing we can do about that. So @gjsjohnmurray go ahead and merge.

Issue #1170, opened 2023-06-20
Debugger loses sync on larger volume output test suite


jpa95:
When running unit tests that output large amount of text (think many tests and assertions in this class) the debugger loses sync (seems to hang) after a short while.
This was reproduced 10 out of 10 times. We do not have this problem on smaller test suites.

Studio v 2022.1.2 Build 574 runs this test suite correctly 100% of the time.

Note that when looking at the test run in the Management portal, the test suite run in VsCode is reported as being successful. So the tests are running but for some reason we are running out of buffer output space or something. 

Test Run #1
46710 bytes output / 352 lines / 279 Assertions / stuck after starting test #14

Test Run #2 (same test case order as Test Run #1)
46708 bytes output / 352 lines / 279 Assertions / stuck after starting test #14

Test Run #3 (different test case order)
49173 bytes output / 368 lines / 279 Assertions / stuck during test #14

Reproduced on : 
- InterSystems 2022.1.2.574.0
- VsCode 1.79.2
- InterSystems Language Server v2.3.5
- InterSystems ObjectScript v2.8.1
- Intersystems Server Manager v3.2.2

Best Regards!

isc-bsaviano:
Hi @jpa95, this looks like a duplicate of #751 but I will try to reproduce again.

jpa95:
Hi @isc-bsaviano, just tested your testcase in #751, and I always get the error message [xmldom warning] at file 166/167. Please see attached logs. Thanks!

[751 logs.zip](https://github.com/intersystems-community/vscode-objectscript/files/11803220/751.logs.zip)

isc-bsaviano:
@jpa95 I think I found the solution to this. I will include the fix in a PR with a handful of other debugger changes that I have queued up.

GendronAC:
Great job ! :) 

![image](https://github.com/intersystems-community/vscode-objectscript/assets/45271725/f1d477fe-a5e9-4616-90a2-feda5ea1a55e)

jpa95:
Thank you @isc-bsaviano !

PR #1258, opened 2023-10-26
Improve handling of debugging WebSocket messages

isc-bsaviano:
The server generally tries to send one "line" of console output per WebSocket "packet" when debugging. However, in cases where an extreme amount of output is written to the console in a short amount of time , the server may send a 32K chunk of many "lines". Currently our debug adapter chokes on that huge packet and throws an error, and the debugger becomes out of sync with the server. This PR fixes that.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,7 +29,7 @@
         "@types/mocha": "^7.0.2",
         "@types/node": "^14.18.0",
         "@types/vscode": "1.75.0",
-        "@types/ws": "^8.5.8",
+        "@types/ws": "8.5.4",
         "@types/xmldom": "^0.1.29",
         "@typescript-eslint/eslint-plugin": "^4.32.0",
         "@typescript-eslint/parser": "^4.32.0",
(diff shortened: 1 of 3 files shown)


gjsjohnmurray (approved):
LGTM though not actually tested

jpa95:
This looks a lot like the issue we had with #1170. The fix provided at that time (#1174) did help some but not completely, we still have large output tests freeze up (the test goes through but the output window freezes at some point). I juste never go around to re-submitting.

isc-bsaviano:
@gjsjohnmurray @jpa95 @GendronAC I fixed the build error so you can now download [this vsix](https://github.com/intersystems-community/vscode-objectscript/suites/17658471703/artifacts/1009853738) to test the change if you want.

gjsjohnmurray (approved):

GendronAC:
I installed it and ran a few hundreds testcase without issue :) #HappyDance :shipit:

jpa95:
Working great on my specific use case as well. Thanks a lot!

PR #1269, opened 2023-11-02
Add support for running and debugging unit tests

isc-bsaviano:
This PR can be tested with the latest 2023.3 preview kit (build 245). The documentation has not been written yet, but here are some bullet points about how this new feature works that I have passed to our docs team:

* The extension supports running tests in client-side or server-side workspace folders.
 * For server-side workspace folders, all test classes in the folder are shown in the Testing view. A root in the Testing view is created for each server-side workspace folder.
 * For client-side workspace folders, you must configure the `objectscript.unitTest.relativeTestRoots` setting to tell the extension where your tests are located. A root in the Testing view is created for each element in this array.
 * When running server-sider test classes, the versions of the classes stored on the server are run, and the tests are re-compiled before they are run.
 * When running client-side tests, the extension loads all test classes from your local workspace folder onto the server, overwriting any versions that may have existed. The extension also loads any autoload files. The autoload feature behavior can be configured by the `objectscript.unitTest.autoload` settings. All files in autoload folders that are in the tree of tests to run, up to the `objectscript.unitTest.relativeTestRoots` value for that Testing root, will be loaded on the server as well. Any documents that were created on the server by the loading will be deleted after the test run is complete. The autoload feature supports loading .cls, .mac, .int and .inc files, as well as legacy XML export files.
 * All client-side test classes and autoload files are sent to the server in a single REST request. This means that you could theoretically surpass the size limit for POST requests set by your web server. I consider this unlikely since this was designed for running small suites of tests during development, and not for loading large suites outside of development.
 * The value of ^UnitTestRoot is not modified during the test run, so whatever value it is currently set to will be used. This is important to note if your tests use that value to load files from the file system during the run.
 * The extension only supports running tests that are in files with the .cls extension. If your tests are in legacy XML export format, you must convert them to .cls files to run them in a client-side workspace folder. If you use server-side editing, you can use the [WebSocket Terminal](https://docs.intersystems.com/components/csp/docbook/DocBook.UI.Page.cls?KEY=GVSCO_debug#GVSCO_debug_websocket) to run the [DebugLoadTestSuite()](https://irisdocs.intersystems.com/irislatest/csp/documatic/%25CSP.Documatic.cls?LIBRARY=%25SYS&CLASSNAME=%25UnitTest.Manager#DebugLoadTestSuite) method in %UnitTest.Manager to load the tests into your server.
[...]

PR #1270, opened 2023-11-03
Use Server Manager's View Container

isc-bsaviano:
The goal of this PR and its dependency is to group all three InterSystems views (Explorer, Projects, Servers) under a single "InterSystems" view container. The end result for the user will be a single view container using the ISC logo containing all three views, instead of the Servers view being under the "InterSystems Tools" view container (toolbox.

Requires https://github.com/intersystems-community/intersystems-servermanager/pull/216


gjsjohnmurray:
Today I have installed this dev VSIX and the Server Manager one (https://github.com/intersystems-community/intersystems-servermanager/pull/216), so they'll start getting some usage here.

Perhaps we should release this as 2.12.0 in recognition of its new dependency on Server Manager, which we are also intending to increment the minor version of (to 3.6.0)

isc-bsaviano:
Thanks for trying this John! I didn't bump the minor version here because I did it in #1269 and I figured that both of these PRs would be in a hypothetical 2.12.0.

isc-rsingh:
@isc-bsaviano with the merged views we lose the "ObjectScript" heading on the container, so I suggest changing the "Explorer" panel to "ObjectScript Explorer". 

image

isc-bsaviano:
@isc-rsingh I don't mind calling it the "InterSystems Explorer" (the view container's name plus the view's name), which is what will be shown if the view is moved from the default view container. I've already alerted Shawn that the docs will need some changes to accommodate this pull request so I don't see this minor name change as a problem.

isc-rsingh (approved):

gjsjohnmurray (approved):

Issue #1273, opened 2023-11-14
debug terminal does not display all output from class


ltjand:
Type: <b>Bug</b>

when passing a json object to a recursive function, it does not 'write' all output to the debug console.  It works when the class method is run from intersystems terminal, but does not seem to work from vscode

Extension version: 2.10.5
VS Code version: Code 1.75.1 (441438abd1ac652551dbe4d408dfcec8a499b8bf, 2023-02-08T21:32:34.589Z)
OS version: Windows_NT x64 10.0.19045
Modes:
Sandboxed: No

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz (8 x 2112)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: disabled_off<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_renderer: enabled_on<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: disabled_off|
|Load (avg)|undefined|
|Memory (System)|31.81GB (11.00GB free)|
|Process Argv|--crash-reporter-id 7d2841b2-426d-4ce1-81dd-0f8726e2c3ac|
|Screen Reader|no|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368:30146709
vsreu685:30147344
python383:30185418
vspor879:30202332
vspor708:30202333
vspor363:30204092
vstes627:30244334
vslsvsres303:30308271
vserr242cf:30382550
pythontb:30283811
vsjup518:30340749
pythonptprofiler:30281270
vshan820:30294714
vstes263:30335439
vscorecescf:30445987
vscod805:30301674
binariesv615:30325510
bridge0708:30335490
bridge0723:30353136
vsaa593:30376534
pythonvs932:30410667
py29gd2263cf:30880073
vscaat:30438848
vsclangdc:30486549
c4g48928:30535728
dsvsc012cf:30540253
pynewext54:30695312
azure-dev_surveyonecf:30548226
2e4cg342:30602488
f6dab269:30613381
showlangstatbar:30737416
pythonfmttext:30731395
[...]

PR #1276, opened 2023-11-27
Harden `TextSearchProvider`

isc-bsaviano:
Our `TextSearchProvider` assumes that the server won't send back junk results. This leads to it choking if it tries to process an invalid document name. This problem can arise due to a server-side bug described [here](https://github.com/intersystems-community/vscode-objectscript/issues/1273#issuecomment-1816587381). This PR hardens the code to ignore clearly invalid document names.


gjsjohnmurray (approved):
LGTM

Issue #1256, opened 2023-10-24
VSCode Plugin Clears File Content When Copying


hsyhhssyy:
As previously mentioned in Issue #1233 , the latest version v2.10.4-beta.7 still exhibits the following issues related to the 'Copy' action:

**Case 1**: When dragging a file with a .cls extension from the Windows Operating System's File Explorer into VSCode's File Explorer, the file's content is automatically cleared. I expect the file to remain untouched and its content unchanged.

![ExplorerDrag](https://github.com/intersystems-community/vscode-objectscript/assets/5977110/f4f3b74b-e677-4909-b709-ab35bfa18031)

**Case 2**: When dragging copy a file (drag while holding Ctrl key) with a .cls extension from one folder in the VSCode File Explorer to another folder, the file's content is automatically cleared. Again, I expect the file to remain untouched and its content unchanged.

![CtrlDrag](https://github.com/intersystems-community/vscode-objectscript/assets/5977110/1079ac39-6c15-4ccd-ac97-ad55b0fe07dd)

**Case 3**: When I perform the COPY command (either using Ctrl+C or the Copy menu item) and paste it into the VSCode File Explorer, the file's content is automatically cleared. As with the previous cases, I expect the file to remain untouched and its content to be preserved.

![CopyPaster](https://github.com/intersystems-community/vscode-objectscript/assets/5977110/252a2869-a472-4b3c-9a9f-6d979faa2bf0)

Are the behaviors described above intentional or are they bugs?

isc-bsaviano:
@hsyhhssyy Sorryb for not responding on your previous issue, I was on vacation. I will try to reproduce this now.

isc-bsaviano:
I was not able to reproduce the "file clearing" bug in any of your three cases using version 2.10.3 (the latest released version) on my Mac. I will try to use a Windows machine in case this is Windows-specific.

hsyhhssyy:
Upon further investigation, it appears that the behavior of this case might be influenced by the specific environment you're connecting to. Depending on whether or not a connection to a server is established, and whether this class is present on that server, the plugin may exhibit varied behaviors. I'll attempt to recreate this in a more controlled environment, like a new VM, to better understand the underlying issue.

gjsjohnmurray:
@isc-bsaviano I think it's significant that in the screen recording the top line of the input file begins with lowercase "class", which the following line will not detect:

[...]

PR #1277, opened 2023-12-06
PR to fix my previous issues.

hsyhhssyy:
The current version of the plugin still has several unresolved issues that I have encountered, detailed as follows:

1. When openning server-side Classes using ObjectScript Explorer, pressing Ctrl+S and select a folder from the current workspace invariably prompts a question about overwriting. This happens because, upon file creation, the plugin automatically inserts a stub class code which often has a) an incorrect Package, and b) invariably triggers an overwrite prompt.
2. The issue of the Class file's Package being altered during movement persists in the latest version of the plugin. Every scenario I reported in my previous case still occurs.
3. When a file is a binary file but has a .cls extension, dragging it between folders in VSCode can corrupt its binary structure.

To thoroughly address these issues, I have submitted this PR. The PR introduces a new VSCode option designed to prevent any automated Code Formatter actions without explicit user permission. Any write operations to code files not executed through the user's ContextMenu will be disabled.

Currently, it only acts on onDidCreateFiles to prevent the automatic addition of PackageNames to code. If further similar behaviors are identified in the future, I plan to extend its control scope. For instance, if an ObjectScript Lint is introduced later, the option would block the automatic application of Lint when saving files with Ctrl+S. Users would need to explicitly apply Lint by right-clicking in the file and selecting FormatCurrentFile.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "vscode-objectscript",
-  "version": "2.10.5-SNAPSHOT",
+  "version": "2.10.6-SNAPSHOT",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "vscode-objectscript",
-      "version": "2.10.5-SNAPSHOT",
+      "version": "2.10.6-SNAPSHOT",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1458,8 +1458,8 @@
           "type": "boolean",
           "default": false
         },
-        "objectscript.disableImplictCodeChange": {
-          "description": "Disable all implict change to any of your code.",
+        "objectscript.disableImplicitCodeChange": {
+          "description": "Disable all implicit changes to your code.",
           "type": "boolean",
           "default": false
         },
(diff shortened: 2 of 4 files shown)


hsyhhssyy:
Related Issue: https://github.com/intersystems-community/vscode-objectscript/issues/1256

isc-bsaviano (changes requested):
@hsyhhssyy Thank for the PR. I've added some suggestions above.

@gjsjohnmurray @isc-rsingh Can you take a look at this as well? I think the language of the new setting needs some work.

hsyhhssyy:
I've accepted these suggestions and committed new code.

gjsjohnmurray (dismissed):

isc-rsingh (dismissed):

isc-bsaviano:
@hsyhhssyy Thanks for making the suggested changes! I'm getting ready to merge this PR, but one of our CI tests is failing due to linting issues. Can you run `npm run lint-fix` in your branch and push the changes that get made?

hsyhhssyy:
I have executed `npm run lint-fix` and submitted a new commit.

isc-bsaviano:
Thanks @hsyhhssyy, I just merged your PR!

PR #1290, opened 2023-12-28
Make server-side search respect Context Lines feature of Search Editor

### LINKED ISSUE #1286: Server-side (isfs) search ignores Context Lines feature of Search Editor
Type: <b>Bug</b>

1. Be using a beta version of the extension, with proposed APIs enabled.
2. Use Server Manager to create a workspace giving isfs access to a namespace.
3. Click 'Search' in Activity Bar and perform a search that returnes some results. in the Side Bar .
4. Click the 'Open in editor' link alongside the search counts.
5. In the Search Editor tab that appears, make sure the 'Toggle Context Lines' button at the right-hand end of the search field line is selected nand that the adjacent number field contains a positive integer.

:bug: no context lines are displayed.

The fault is that this function ignores `options.beforeContext` and `options.afterContext`.

https://github.com/intersystems-community/vscode-objectscript/blob/66f0f37e156eda4bde4831c846db9f458f29813b/src/providers/FileSystemProvider/TextSearchProvider.ts#L307-L308

Extension version: 2.10.6-dev.1270
VS Code version: Code 1.85.1 (0ee08df0cf4527e40edc9aa28f4b5bd38bbff2b2, 2023-12-13T09:49:37.021Z)
OS version: Windows_NT x64 10.0.22621
Modes:

gjsjohnmurray:
This PR fixes #1286


isc-bsaviano:
@gjsjohnmurray I think the CI keeps failing because we need to update `@vscode/test-electron` to the latest version. Can you do that in this branch so I can test this change?

gjsjohnmurray:
@isc-bsaviano done, and CI now happy.

isc-bsaviano:
Thanks for that @gjsjohnmurray. I gave this a try and while the code looks fine, I think it's performing much worse than before when there are many matches. I think that the issue is with your `previewFrom` and `previewTo` calculations. I think you need parentheses, so `line - options.beforeContext || 0` becomes `line - (options.beforeContext || 0)`.

gjsjohnmurray:
Good catch! I have pushed the fix.

isc-bsaviano:
You fix is better, but I'm still seeing some slowness. The slowness occurs when I start with an empty search box (I'm using the regular search view, not the editor) and type a string to search. Using the current beta, the previous searches are cancelled very quickly when a new character is typed and the final search is shown quickly once typing is stopped. Using this dev vsix, the old searches are not being cancelled fast enough so there is lag when new characters are typed where the search results shown are for the older search. I'm not sure how this could be caused by your change since the context lines options will be `undefined`.

isc-bsaviano:
@gjsjohnmurray Once again, I really like your edit but the search still gets "stuck" on the previous query while you type. Maybe we can add more checks of `token.isCancellationRequested`? I would add a check at the top of `searchMatchToLine` and the `forEach` callback for the `Set`. Sorry for not being able to provide better reproduction details, I just want to avoid performance degradation for the most common search case.

gjsjohnmurray:
@isc-bsaviano please re-test

isc-bsaviano:
@gjsjohnmurray Just tested again and am seeing the same behavior. Here's two screen recordings, one with beta.3 and the other with your most recent dev.1290:


https://github.com/intersystems-community/vscode-objectscript/assets/44776135/26f6c782-3a70-43b1-8fb4-e04d0e15211b

https://github.com/intersystems-community/vscode-objectscript/assets/44776135/7502c90d-fcfb-4105-bcc8-6ad67c191bea

gjsjohnmurray:
What is the $ZV of the server you are connecting to?

isc-bsaviano:
It's an internal kit. I can try again with a 2023.3 community kit.

gjsjohnmurray:
Here's what I get with VS Code 1.85.1 Windows connecting to a local Docker instance of the containers.intersystems.com/intersystems/iris-community:2023.3 image

junk

isc-bsaviano:
Interesting. I'm on an M1 Mac, which appears to be the only difference in our setup.

gjsjohnmurray:
And strange that your dev.1290 recording ends with "0 results in 1 file".

isc-bsaviano:
Oh, that's because I cut it off a fraction of a second early. It eventually returned the correct results. Maybe that's showing up because context was sent before the match even though there shouldn't be any in this case. I don't see how that could happen from your code though.

isc-bsaviano:
@gjsjohnmurray I copied your changes into one of my branches and in "Extension Development Host" mode everything looked fine. I'm going to approve this and assume that the issue is unique to me. If we see any complaints from the field we can take a closer look then.

isc-bsaviano (dismissed):

isc-bsaviano (approved):

PR #1300, opened 2024-01-18
Remove `glob` as a dependency

isc-bsaviano:
This PR removed `glob` as a dependency, replacing it with `vscode.workspace.findFiles()`. `glob`is still a dev depedency.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,7 +24,6 @@
         "ws": "^8.14.2"
       },
       "devDependencies": {
-        "@types/glob": "^7.1.2",
         "@types/istextorbinary": "2.3.1",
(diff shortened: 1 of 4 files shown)


isc-rsingh:
My only comment is, fewer dependencies are usually better, especially when the library depends on the efforts of a single developer (even though [glob](https://github.com/isaacs/node-glob) seems to be rock solid and heavily used with 8.3k stars).

isc-rsingh (approved):

isc-bsaviano:
@isc-rsingh My motivation for doing this is that VS Code provides a handy findFiles API. There wasn't any need for an external library to do that. I don't have any concerns about glob's long term viability.

PR #1302, opened 2024-01-19
Add testing link to new KPIs

isc-bsaviano:
IRIS BI KPI classes can be tested by opening them as a web page. This PR adds a link that can be opened in a browser so the testing page is easily accessible.


isc-rsingh (approved):

gjsjohnmurray (commented):
The link will only be valid on the server the class was created on, and in the namespace it was created in. Where source control or similar is being used to promote code between environments and/or namespaces to test, stage and deploy it the URL will become stale.

Did you consider implementing a CodeLens? Yes, this would only offer the testing launch when the class is viewed inside VS Code. But it would work correctly as the class moves.

PR #1313, opened 2024-02-13
Fire source control hooks when creating/opening/editing/deleting projects

### LINKED ISSUE #1312: Fire source control hooks when creating/opening/editing/deleting projects
We use a custom source control solution enabled by ISC Source Control hooks to store our ObjectScript source code in an external repository. For the purposes of this, we use project files that export as manifests, and each class or macro is tracked separately. The sample contents of a file:

```
<?xml version="1.0" encoding="UTF-8"?>
<Export generator="IRIS" version="26">
<Project name="Sample_Project">
  <Items>
    <ProjectItem name="MyPackage.SampleClass" type="CLS"></ProjectItem>
    <ProjectItem name="MyPackage.AnotherClass" type="CLS"></ProjectItem>
    <ProjectItem name="SampleINC.INC" type="MAC"></ProjectItem>
  </Items>
</Project>
</Export>
```

In Studio, these hooks generate files after saving a project after adding/removing project artifacts.

Without modification of the code we use to export files, we haven't had issues with any of the source control hooks except for the exporting of these PRJ files.

We are finding this is not the case in Visual Studio Code.

I reached out directly to Brett Saviano, who noted:

_We use SQL for managing projects because the Atelier API doesn't support retrieving the contents of a PRJ file. I can look into manually firing source control events before/after our SQL queries if that would resolve your issue._

Is this a workflow that can be supported through this extension?

isc-bsaviano:
This PR fixes #1312

I think this is the best I can do with "other" actions to approximate what happens when a file is opened via the Atelier GET /doc/ API. @gjsjohnmurray I'd really appreciate your review on this since since you're a server-side source control expert.

### DIFF
--- a/src/commands/project.ts
+++ b/src/commands/project.ts
@@ -12,6 +12,7 @@ import { isCSPFile } from "../providers/FileSystemProvider/FileSystemProvider";
 import { notNull, outputChannel } from "../utils";
 import { pickServerAndNamespace } from "./addServerNamespaceToWorkspace";
 import { exportList } from "./export";
+import { OtherStudioAction, StudioActions } from "./studio";
 
 export interface ProjectItem {
   Name: string;
@@ -137,6 +138,20 @@ export async function createProject(node: NodeBase | undefined, api?: AtelierAPI
         return;
       }
 
+      // Technically a project is a "document", so tell the server that we created it
+      try {
+        await new StudioActions().fireProjectUserAction(api, name, OtherStudioAction.CreatedNewDocument);
+        await new StudioActions().fireProjectUserAction(api, name, OtherStudioAction.FirstTimeDocumentSave);
+      } catch (error) {
+        let message = `Source control actions failed for project '${name}'.`;
+        if (error && error.errorText && error.errorText !== "") {
+          outputChannel.appendLine("\n" + error.errorText);
+          outputChannel.show(true);
+          message += " Check 'ObjectScript' output channel for details.";
+        }
+        return vscode.window.showErrorMessage(message, "Dismiss");
+      }
+
       // Refresh the explorer
       projectsExplorerProvider.refresh();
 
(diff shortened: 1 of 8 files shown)


src/commands/project.ts
```diff
@@ -137,6 +138,20 @@ export async function createProject(node: NodeBase | undefined, api?: AtelierAPI
         return;
       }
 
+      // Technically a project is a "document", so tell the server that we created it
+      try {
+        await new StudioActions().fireProjectUserAction(api, name, OtherStudioAction.CreatedNewDocument);
```
gjsjohnmurray, line 143: Do we need two new StudioActions objects here or can we create one into a const and use it twice?

gjsjohnmurray (changes requested):
Looks good as code, with only one minor comment made. I haven't tested it but I think the person who raised the issue is better placed to do that.

isc-bsaviano:
Thanks for the review @gjsjohnmurray! I've added your suggestion as well as a few more changes.

gjsjohnmurray (dismissed):
LGTM but not tested

isc-rsingh (approved):

gjsjohnmurray (approved):

trevorfstrong:
Hi @isc-bsaviano we have always wanted the exposure of the current project opened for the current document to be passed to VSC studio Source Control hooks. Sounds like this might do the trick.
How can we update the plugin to trial this with our studio hooks?

isc-bsaviano:
@trevorfstrong I just re-read your request and I think it may be unrelated to this PR. This PR fires OtherStudioActions for .PRJ files. Is that what you're asking for, or do you mean something else?

isc-bsaviano:
@trevorfstrong Something like that is going to be very difficult to implement in VS Code. VS Code doesn't really "open" projects, it just treats them as a SQL table that defines a group of documents. Since users can have multiple projects "open" at once (via the multi-root workspace feature) and REST connections can go through a number of CSP jobs, there isn't a good way to persist that information on the server.

trevorfstrong:
ok, but if you passed the project in this screenshot when you "access" a document that is linked to the project, we can persist it on the back end as the last project "touched" by vsc.
image

isc-bsaviano:
I understand. It would be possible to pass the project with the REST requests somehow and "open" it for the duration that the request is processed so any source control hooks see it. That would require a bunch of IRIS-side changes though. I would need to discuss that internally before drawing up a spec or committing to doing that work.

trevorfstrong:
Thanks for your attention on this subject Brett.

isc-bsaviano:
Thanks for bringing this to my attention! I will add this to my notes for when we have a larger discussion about projects support in VS Code.

trevorfstrong:
Hi Brett.
Looks like the firing of project actions in the latest code is always triggered before items for source control actions, so i believe i can work with that. No need for any further mods

trevorfstrong:
Hi again.
When we have multiple projects associated with the workspace and we select a file that is in a specific project, you fire the src ctrl actions for all projects in the workspace then that selected file, but the projects are simply looped thru in the order they are presented in the workspace when fired, so is there any way the project src ctrl action that the current file selected is in be fired last? e.g. before the currently selected file?

trevorfstrong:
image

PR #1314, opened 2024-02-15
Server-side source control improvements

isc-bsaviano:
* Don't show source control menu options for root directories or directories in Web Application workspace folders.
* Don't check if source control is enabled before firing Other Studio Actions. These queries will return nothing if no source control class is enabled, so the check is just adding an extra query for no benefit. 
* Move the progress notification to the window's Status Bar. This is less distracting IMO.
* Preserve focus when showing errors in the Output channel.
* Write queries to the Output channel as well as user actions when `objectscript.studioActionDebugOutput` is `true`.
* Add type for UserAction.

I won't merge this until #1313 is in

### DIFF
--- a/src/commands/studio.ts
+++ b/src/commands/studio.ts
@@ -9,6 +9,7 @@ import { RoutineNode } from "../explorer/models/routineNode";
 import { importAndCompile } from "./compile";
 import { ProjectNode } from "../explorer/models/projectNode";
 import { openCustomEditors } from "../providers/RuleEditorProvider";
+import { UserAction } from "../api/atelier";
 
 export let documentBeingProcessed: vscode.TextDocument = null;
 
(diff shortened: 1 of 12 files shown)


src/commands/studio.ts
```diff
@@ -295,94 +295,85 @@ export class StudioActions {
       ? [type.toString(), action.id, this.name, answer, msg]
       : [type.toString(), action.id, this.name, selectedText];
 
-    return vscode.window.withProgress(
-      {
-        cancellable: false,
-        location: vscode.ProgressLocation.Notification,
```
gjsjohnmurray, line 287: Rather than removing the progress indicator did you experiment with changing its location to SourceControl or Window?

gjsjohnmurray (changes requested):

src/commands/studio.ts
isc-bsaviano: I did try the SourceControl location. It was much less invasive (just a blue clock badge on the source control view icon). We don't actually populate that view though so I didn't think it made sense to call any attention to it. I didn't try the Window location. I think that most actions are very "in your face" (showing a modal dialog or web page etc) so we don't need another visual indicator.

isc-bsaviano: Here's what the Window progress location looks like:

Screenshot 2024-02-20 at 7 46 43 AM

I think this is fine as a compromise.

gjsjohnmurray (dismissed):

PR #1317, opened 2024-02-22
Better handling of generated INT routines in project folders

isc-bsaviano:
Generated INT routines shouldn't have their last dotted piece before the extension (e.g. the `1` in `User.Test.1.int`) split into a folder.


isc-bsaviano:
@gjsjohnmurray The Projects Explorer and regular isfs both treat generated files as described in this PR's description, although the regular isfs change to enable it was a server-side change so I'm sure what versions it's in. This PR shouldn't affect many people, but it brings all of our explorers in line.

isc-rsingh (approved):

PR #1319, opened 2024-02-23
Reduce firing of `OpenedDocument` action for projects

isc-bsaviano:
I've reduced the `OpenedDocument` calls to what I consider the bare minimum (when expanding the root of a project ISFS folder, expanding a root of the Projects Explorer, or modifying a project). See https://github.com/intersystems-community/vscode-objectscript/issues/1312#issuecomment-1954709482 for motivation.


gjsjohnmurray (approved):
Looks fine to me. This isn't relevant to Deltanji, so go ahead if the customer is happy with the change.

PR #1321, opened 2024-02-23
Refresh ObjectScript Explorer files when they are re-opened

### LINKED ISSUE #1320: InterSystems Explorer doesn't automatically refresh a class changed from outside VS Code
Consider a class that's been opened in VS Code from the InterSystems Explorer view. It may or may not have been exported to the local file system for editing. Now, change the class using a different interface. For example:

- If the class is a persistent class, run ALTER TABLE ADD COLUMN from the SQL section of the Portal.
- If the class is a DTL class, add something to the DTL in the Portal.

Both of these actions change the class on the server. However, closing and reopening the class **doesn't** show the updated class definition. I don't think "Refresh Explorer" is designed to do this, but it doesn't show the updated class either. It's as if VS Code has its own local copy and doesn't know that the server version has been changed.

From the InterSystems Explorer view, the following actions **do** update the class:
- Exporting the class from the server.
- Compiling the class (if the "u" compiler option is **not** set).

isc-bsaviano:
This PR fixes #1320

### DIFF
--- a/src/explorer/explorer.ts
+++ b/src/explorer/explorer.ts
@@ -2,7 +2,7 @@ import * as vscode from "vscode";
 import { NodeBase } from "./models/nodeBase";
 
 import { AtelierAPI } from "../api";
-import { config, OBJECTSCRIPT_FILE_SCHEMA, projectsExplorerProvider } from "../extension";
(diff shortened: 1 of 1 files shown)


isc-rsingh (approved):

gjsjohnmurray (approved):
Looks OK, so if the OP is happy with testing let's merge it.

PR #1329, opened 2024-03-04
Support opening links from Studio extension pages

### LINKED ISSUE #1328: Add a way for an extension page to open an external link in an external browser
If I have a CSP page launched in a Studio extension, and there's a link in there to an external site, nothing will happen. You can see this with %Studio.SourceControl.ISC, Source Control > Show CCR Controls, in a namespace configured for CCR - the "Create New CCR" link just doesn't work.

Solution could be something like what is described in https://stackoverflow.com/a/68993592 - allow the iframe to pass a message back to the parent, and handle it on the VSCode side to trigger opening an external browser.

@isc-bsaviano I'm happy to use this as an excuse to train my team a bit on contributing. ;)

isc-bsaviano:
This PR fixes #1328 

@gjsjohnmurray @isc-rsingh Should I do any input validation on the VS Code side before attempting the `openExternal()`? That failing should be harmless since it's a Thenable and we aren't awaiting it. I doubt a Studio extension page is going to have links to spam web pages.

### DIFF
--- a/src/commands/studio.ts
+++ b/src/commands/studio.ts
@@ -155,7 +155,9 @@ export class StudioActions {
               answer = "1";
               panel.dispose();
             } else if (typeof message.href == "string") {
-              vscode.env.openExternal(vscode.Uri.parse(message.href));
+              const linkUri = vscode.Uri.parse(message.href);
+              // Only open http(s) links
+              if (/^https?$/.test(linkUri.scheme)) vscode.env.openExternal(linkUri);
             }
           });
           panel.onDidDispose(() => resolve(answer));
--- a/src/commands/studio.ts
+++ b/src/commands/studio.ts
@@ -154,6 +154,8 @@ export class StudioActions {
             if (message.result && message.result === "done") {
               answer = "1";
               panel.dispose();
+            } else if (typeof message.href == "string") {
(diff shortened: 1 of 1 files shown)


gjsjohnmurray:
Maybe prudent to restrict this to http(s) only, at least until someone complains.

isc-egabhart:
I took the beta made from this commit for a spin by dropping the file into my VS Code extensions and restarting. Unfortunately, I still observe the same behavior when after adding the onClick event to the "create new CCR" link on the landing page of the CCR Controls:
<A href="#($$CCRServer^%buildccr)#/ccr/create.csp?SiteCode=#($G(^SYS("SourceControl","ItemSetConfig","Org")))#&SystemCode=#($G(^SYS("SourceControl","ItemSetConfig","System")))#&BypassPREP=1" rel="noopener noreferrer" target='_blank' onClick="window.parent.postMessage(this.href, '*')">[create new CCR]</A>
Am I missing something so I can make use of this feature?

isc-bsaviano:
@isc-egabhart Yeah, your message is a little off. Try this instead:

`window.parent.postMessage({ href: this.href }, '*')`

isc-egabhart:
@isc-bsaviano Thanks so much! Works like a charm!

PR #1330, opened 2024-03-06
Source control useractions for projects ignored `"objectscript.serverSourceControl.disableOtherActionTriggers": true` setting

gjsjohnmurray:
Fixes an issue reported by a Deltanji user, caused by the project-level source control calls added in 2.12.2.

The setting previously added to allow Deltanji sites to suppress user action calls was not being respected.


isc-bsaviano (approved):

isc-bsaviano:
I don't have a planned date but I also don't have any work in the pipeline that a release needs to wait for. How badly does the customer need a non-beta version?

gjsjohnmurray:
They tell me there's no rush, so let's leave it a while.

PR #1332, opened 2024-03-13
Update Test Explorer when test classes are renamed

isc-bsaviano:
I assumed that renaming files triggered both a "create" event on the new URI and a "delete" event on the old URI, but that wasn't true. This PR adds a listener for the "rename" event as well so the Test Explorer tree is always accurate.


gjsjohnmurray (approved):
LGTM but I didn't test.

PR #1334, opened 2024-03-18
Only remove blank line ending for web application files

### LINKED ISSUE #1333: Blank Line Endings in Routines Lost on Save
Hi,

We have spotted what looks like a fault with blank line endings when saving routines in VS Code. If the routine ends with a blank line, the blank line is removed by the time it is sent to the PutDoc Atelier API.

Saving a routine which looks like this:

image

Ends up looking like this after one save:
image

And after multiple saves:
image

It looks like the code responsible is here: https://github.com/intersystems-community/vscode-objectscript/blob/0a446be77a012fb2e99f9bf2d7d18a91ebf32ce2/src/commands/compile.ts#L102C8-L102C102

`

    content = file.content.split(/\r?\n/);

    // Avoid appending a blank line on every save, which would cause a web app file to grow each time
    if (content.length > 1 && content[content.length - 1] === "") {
      content.pop();
    }

`

The comment reads as though the string.split is adding a blank line, and that the code below it is removing it. I can't imagine why doing a string.split would add a new line. The issue this causes us is that some of our routines end with a blank line, and when developers simply import and compile one of these routines the developer is left wondering why the file is picked up as an unstaged change.  

Does anyone know why this behaviour is necessary?

Thanks

Mat

isc-bsaviano:
This PR fixes #1333

### DIFF
--- a/src/commands/compile.ts
+++ b/src/commands/compile.ts
@@ -100,7 +100,7 @@ async function importFile(
     content = file.content.split(/\r?\n/);
 
     // Avoid appending a blank line on every save, which would cause a web app file to grow each time
-    if (content.length > 1 && content[content.length - 1] === "") {
+    if (file.name.includes("/") && content.length > 1 && content[content.length - 1] == "") {
       content.pop();
     }
   } else {


mathewrimmingtoncds:
Hi, thanks for making this code change so quickly. Is there a plan to publish a release soon?

isc-bsaviano:
@mathewrimmingtoncds Happy to help! I'm going to wait for #1332 to be merged at the very least. There also is a core VS Code issue affecting this extension that I may need to implement a workaround for as well. If yuu need this fix right now you cna download the latest beta version of the extension from the releases page.

PR #1339, opened 2024-03-28
Remove `objectscript.ignoreInstallServerManager` setting

isc-bsaviano:
This setting is no longer required because this extension depends on the Server Manager in its package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -1477,11 +1477,6 @@
           "default": false,
           "description": "Output in JSON format the action that VS Code should perform as requested by the server."
         },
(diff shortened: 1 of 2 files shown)


isc-bsaviano:
You're right, none of that is necessary! I'll remove it. The `ignoreInstallLanguageServer` setting is undocumented so I figured it was safe to remove. However, I think I should keep it. There are a few platforms that the Language Server doesn't support.

isc-rsingh (approved):

gjsjohnmurray (approved):
LGTM

PR #1351, opened 2024-04-19
Don't prompt for Server Manager credentials when using minimal security

isc-bsaviano:
This PR brings this extension in line with the Language Server. @gjsjohnmurray I won't merge this without your approval.


isc-bsaviano:
@gjsjohnmurray Any comments on this or #1353?

gjsjohnmurray (approved):
Change makes sense, but I'm not 100% clear what effect it has from a user perspective.

isc-bsaviano:
A user reported that they couldn't connect using no username or password without getting prompted to get credentials from the server manager's authentication provider. This change doesn't trigger the authentication provider in cases where the user is trying to use no credentials.

PR #1352, opened 2024-04-22
Update CI

isc-bsaviano:


isc-bsaviano:
Do we really need to build on all three platforms? Action `coactions/setup-xvfb@v1` hasn't been updated to Node 20, but xvfb is already installed in the latest Ubuntu runners. Also, The Windows build started failing and updating the actions didn't help.

isc-rsingh:
I haven't looked into what these checks do, but running successfully on Mac and Windows is key -- and marginally Ubuntu (as a proxy for all Linuxes). Are you saying running these checks on all these platforms aren't helping us ensure that?

isc-bsaviano:
Yes, exactly. This extension is pure JavaScript and therefore is platform independent. Testing on one OS (Ubuntu) is good enough.

gjsjohnmurray:
@daimor ought to know, but I guess one reason for running the CI on the different platforms is so that the tests happen on them. But seeing as no actual tests have been written yet we're unlikely to be losing anything by reducing the CI to a single platform.

isc-bsaviano:
Yeah, that's probably the reason Dmitry implemented the CI the way he did.

isc-bsaviano:
I updated the build CI to only run on Ubuntu. It succeeded, and there are no more deprecation warnings.

gjsjohnmurray (approved):

PR #1366, opened 2024-05-20
`WorkspaceSymbolProvider` improvements

isc-bsaviano:
- Now includes classes as well as class members
- Only shows results from classes actually in the isfs folder (respects query params, doesn’t show anything for csp folders)
- Shows results even if no query is specified
- Uses local URI if present for client-side editing
- Resolves location correctly for delimited identifiers
- Checks cancellation token

### DIFF
--- a/src/providers/WorkspaceSymbolProvider.ts
+++ b/src/providers/WorkspaceSymbolProvider.ts
@@ -1,73 +1,55 @@
 import * as vscode from "vscode";
 import { AtelierAPI } from "../api";
 import { DocumentContentProvider } from "./DocumentContentProvider";
+import { filesystemSchemas } from "../extension";
+import { fileSpecFromURI } from "../utils/FileProviderUtil";
 
 export class WorkspaceSymbolProvider implements vscode.WorkspaceSymbolProvider {
-  private sql: string =
-    "SELECT * FROM (" +
-    "SELECT Name, Parent->ID AS Parent, 'method' AS Type FROM %Dictionary.MethodDefinition" +
-    " UNION ALL %PARALLEL " +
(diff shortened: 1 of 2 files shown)


src/providers/WorkspaceSymbolProvider.ts
```diff
@@ -1,73 +1,55 @@
...
-    " UNION ALL %PARALLEL " +
-    "SELECT Name, Parent->ID AS Parent, 'query' AS Type FROM %Dictionary.QueryDefinition" +
-    " UNION ALL %PARALLEL " +
-    "SELECT Name, Parent->ID AS Parent, 'trigger' AS Type FROM %Dictionary.TriggerDefinition" +
-    " UNION ALL %PARALLEL " +
-    "SELECT Name, Parent->ID AS Parent, 'storage' AS Type FROM %Dictionary.StorageDefinition" +
-    " UNION ALL %PARALLEL " +
-    "SELECT Name, Parent->ID AS Parent, 'projection' AS Type FROM %Dictionary.ProjectionDefinition" +
-    ") WHERE %SQLUPPER Name %MATCHES ?";
+  private readonly _sqlPreifx: string =
```
isc-rsingh, line 8: This is just cosmetic, but @isc-bsaviano did you want this variable to be called `_sqlPrefix` (not Preifx)?

isc-bsaviano: Yeah, this is intentional. I chose `_sqlPrefix` to mean "the start of the query that is always used".

isc-rsingh: I was just saying you wrote `ifx` in the code instead of `fix`

isc-bsaviano: Good catch, I totally missed that!

isc-rsingh (approved):

PR #1384, opened 2024-06-20
fixed header name

### LINKED ISSUE #1383: Atelier API behind nginx
When IRIS is working behind webgateway on Nginx, every change of any code says, that file is modified on the server, while it's not

daimor:
By standard header name should be `if-none-match` not `if_none_match`

This PR fixes #1383


isc-bsaviano:
@daimor The underscore is perfectly valid per RFC 822. The header dropping seems like an nginx-specific behavior. Regardless, I verified using Apache and IIS that changing the header to dashes works fine, so this should be harmless.

isc-bsaviano (approved):

PR #1387, opened 2024-06-25
XML Import/Export commands shouldn't assume a server connection in a multi-root workspace

isc-bsaviano:
This PR addresses some confusing behavior reported by an internal dev. They had a multi-root workspace connected to %SYS and USER on the same server. They had a document from %SYS in the active editor tab, but wanted to import an XML file into USER. The command defaulted to use the %SYS connection, so the file got imported into the wrong namespace. I think it's a better user experience to always ask for the connection if you're in a multi-root workspace.

### DIFF
--- a/src/commands/compile.ts
+++ b/src/commands/compile.ts
@@ -812,25 +812,22 @@ interface XMLQuickPickItem extends vscode.QuickPickItem {
 
 export async function importXMLFiles(): Promise<any> {
   try {
-    // Use the server connection from the active document if possible
-    let connectionUri = currentFile()?.uri;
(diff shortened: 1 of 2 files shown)


gjsjohnmurray (commented):
Does this change have any impact on people who work with the client-side editing paradigm?

isc-bsaviano:
No, the behavior is the same for client-side and server-side workspaces. The only change is that the command won't use the server connection of the active text editor by default.

gjsjohnmurray (approved):
Approved on the basis of code-reading only.

PR #1390, opened 2024-06-28
Hide file explorer context commands when multiple items are selected

isc-bsaviano:
Almost all of the commands that we contribute to the file explorer context menu are only apply to a single item. Therefore, it's confusing if the menu options appear when multiple items are selected since the user may not be able to tell which item the command is run with. I propose that we just hide the menu options of mulitselect is active.

### DIFF
--- a/package.json
+++ b/package.json
@@ -360,12 +360,7 @@
       "view/item/context": [
         {
           "command": "vscode-objectscript.explorer.export",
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (approved):
Good idea

isc-rsingh (approved):

PR #1398, opened 2024-07-08
Add command for opening InterSystems documents

### LINKED ISSUE #1379: Server-Side editing Open items as in Studio
Using Studio is possible to open ANY "file" (class, routine, include, csp, csr, etc.) that is "available" (local or via mapping) in the current namespace.
I don't think this is possible using VS Code, or at least, if possible, it's not as easy and simple.

For example, if I need to open a % class and I've copied the classname in clipboard from documentation (or I type it in), in Studio I just press ctrl-O, ctrl-V (or type the name) and press enter and the class is opened.
This is possible not only for classes but for any other "file", for example if I want to open %occInclude.inc (a very popular include file) I can easily do it in Studio.
In fact, in Studio using the File Open dialog I can select the file type I want to open and choose/browse from a list, this is also very useful if you don't remember the exact spelling/case of an item and provide autocompletion as I type as well.
Another example, in Studio you can open the class and/or the routine(s) generated for an SQL statement, you get the class or routine name from Management Portal and in Studio you just press ctrl-O and paste the class or routine name and you have it opened.

Even for items that are part of the current virtual folder structure I did not find an easy way to open a class without browsing the virtual folder structure. If I have copied in clipboard a classname, say "HS.FHIR.DTL.SDA3.vR4.Address.Address.cls", how can I open it immediately? Browsing the folder structure is a pain!
I've tried using [the new "Quick Open" feature](https://code.visualstudio.com/docs/getstarted/tips-and-tricks#_quick-open) but that feature works only for "file" name, in my example the file name is "Address.cls" and I get a list of more then 20 files to choose from, many with similar packege name/path! Quick is an opinion.....

As a side note, to me the fact that for the class HS.FHIR.DTL.SDA3.vR4.Address.Address.cls the file name is actually Address.cls is...well...debatable, I consider the package as integral part of a classname. But maybe that my opinion.

IMHO in VS Code you should be able to open the same "items" I used to open in Studio. Hopefully as easily as in Studio.

I'm using v2.12.6-beta.1 with proposed API enabled.

isc-bsaviano:
This PR fixes #1379. The new `Open InterSystems Document...` command can be opened from the command palette and it provides a UI for browsing all files in the connected server-namespace. The user can either pick an item in the list, or type text into the filter box and press "Enter" to submit that. The UI should function very similarity to the built-in simple file dialog. I would like to assign a keybinding to this command, but I couldn't find a good one so I'm open to suggestions. 

output

### DIFF
--- a/package.json
+++ b/package.json
@@ -359,6 +359,10 @@
           "command": "vscode-objectscript.explorer.project.openOtherServerNs",
           "when": "view == ObjectScriptProjectsExplorer && vscode-objectscript.projectsExplorerRootCount > 0",
           "group": "navigation"
+        },
+        {
+          "command": "vscode-objectscript.openISCDocument",
+          "when": "view == workbench.explorer.fileView && vscode-objectscript.connectActive && workspaceFolderCount != 0"
         }
       ],
       "view/item/context": [
--- a/package.json
+++ b/package.json
@@ -338,6 +338,10 @@
         {
           "command": "vscode-objectscript.compileIsfs",
           "when": "false"
+        },
+        {
+          "command": "vscode-objectscript.openISCDocument",
+          "when": "vscode-objectscript.connectActive && workspaceFolderCount != 0"
         }
       ],
       "view/title": [
@@ -1179,6 +1183,11 @@
         "category": "ObjectScript",
         "command": "vscode-objectscript.compileIsfs",
         "title": "Compile"
+      },
+      {
+        "category": "ObjectScript",
+        "command": "vscode-objectscript.openISCDocument",
+        "title": "Open InterSystems Document..."
       }
     ],
     "keybindings": [
--- a/package.json
+++ b/package.json
@@ -362,7 +362,8 @@
         },
         {
           "command": "vscode-objectscript.openISCDocument",
-          "when": "view == workbench.explorer.fileView && vscode-objectscript.connectActive && workspaceFolderCount != 0"
+          "when": "view == workbench.explorer.fileView && vscode-objectscript.connectActive && workspaceFolderCount != 0",
+          "group": "navigation"
         }
       ],
       "view/item/context": [
@@ -1191,7 +1192,8 @@
       {
         "category": "ObjectScript",
         "command": "vscode-objectscript.openISCDocument",
-        "title": "Open InterSystems Document..."
+        "title": "Open InterSystems Document...",
+        "icon": "$(go-to-file)"
       }
     ],
     "keybindings": [
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -314,7 +314,7 @@ export class AtelierAPI {
     const target = `${username}@${host}:${port}`;
     let auth: Promise<any>;
     let authRequest = authRequestMap.get(target);
-    if (cookies.length || method === "HEAD") {
+    if (cookies.length || (method === "HEAD" && !originalPath)) {
       auth = Promise.resolve(cookies);
 
       // Only send basic authorization if username and password specified (including blank, for unauthenticated access)
@@ -367,8 +367,16 @@ export class AtelierAPI {
       }
       await this.updateCookies(response.headers.raw()["set-cookie"] || []);
       if (method === "HEAD") {
-        authRequestMap.delete(target);
-        return this.cookies;
+        if (!originalPath) {
+          authRequestMap.delete(target);
+          return this.cookies;
+        } else if (response.status >= 400) {
+          // The HEAD /doc request errored out
+          throw { statusCode: response.status, message: response.statusText, errorText: "" };
+        } else {
+          // The HEAD /doc request succeeded
+          return response.headers.get("ETAG");
+        }
       }
 
       // Not Modified
@@ -490,6 +498,11 @@ export class AtelierAPI {
       filter,
       generated,
     });
+  }
+
+  // api v1+
+  public headDoc(name: string): Promise<string> {
+    return this.request(1, "HEAD", `${this.ns}/doc/${name}`);
   }
 
   // api v1+
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -139,6 +139,7 @@ import { RESTDebugPanel } from "./commands/restDebugPanel";
 import { modifyWsFolder } from "./commands/addServerNamespaceToWorkspace";
 import { WebSocketTerminalProfileProvider, launchWebSocketTerminal } from "./commands/webSocketTerminal";
 import { setUpTestController } from "./commands/unitTest";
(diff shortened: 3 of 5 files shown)


isc-rsingh (dismissed):

isc-bsaviano:
@isc-rsingh @gjsjohnmurray How do you feel about using a chord ask the keybinding for this? `Ctrl+K Alt+O` (`Cmd+K Alt+O` on Mac) is available. We could also overload `Ctrl+K Shift+O` (`Cmd+K Shift+O` on Mac), which is only use din the diff editor. This is the best that I could come up with ending with `O`.

isc-rsingh:
How about `Ctrl+K Ctrl+O` (Mac: `Cmd+K Cmd+O`) which seems easy to type and available.

isc-bsaviano:
That's taken on Windows

isc-rsingh:
Ugh. Now I see the semantics are different on Windows and Mac. On Mac, Cmd+O does both the open file and open folder semantics. While on Windows there are 2 separate keybindings, Ctrl+O for open file and the chord `Ctrl+K Ctrl+O` for open folder. I think the best available option is the chord `Ctrl+K Ctrl+Shift+O` and Mac `Cmd+K Cmd+Shift+O`?

isc-bsaviano:
I think `Ctrl+K Alt+O` is better because it's fewer keys but I'm not sure any of these are natural/easy to use.

isc-rsingh:
My suggestion was based on the convenience of not lifting the finger off of `Ctrl` (or `Cmd`) when typing the 2nd part of the chord, which is why I preferred it over a chord that changes all the finger positions, but its not a huge difference.

isc-bsaviano:
That's a good point. Let's see what John thinks.

gjsjohnmurray:
Adding keybindings to new versions of existing extensions brings the risk that some current users will experience problems because the combo we choose clashes with another extension or a keybinding they set up.

I'd be inclined not to set one.

isc-bsaviano:
I'm fine with not setting one

isc-bsaviano:
If you're both happy with this I'll merge it

gjsjohnmurray:
I shan't have time to review / try this before tomorrow, so go ahead if you want.

isc-bsaviano:
No problem John, I'll wait for you review

gjsjohnmurray:
How about adding the command to the Explorer view title menu, perhaps with `go-to-file` as its icon?

gjsjohnmurray:
Mapped .inc documents are not being hidden. To reproduce this, be connected to a namespace such as USER, set toggles to show system documents and to hide mapped ones. Observe that some %-packages still get listed. Drill into one and discover that you are being offered .inc documents despite them being mapped here from another database.

gjsjohnmurray:
It would be nice if the three toggle buttons remembered their state between invocations of the selector, at least for the duration of my session, and perhaps persist per workspace.

It would also be nice if I could see the state of each. AFAIK the only way of doing this at the moment might be to create 6 of our own icons, 3 to represent "on" and 3 "off", then manipulate the `buttons` array. Perhaps the "on" icons would have a solid outline.

Alternatively I could have a go at a PR for https://github.com/microsoft/vscode/issues/185356, or better still my suggestion at https://github.com/microsoft/vscode/issues/221397#issuecomment-2225357158 gets accepted.

isc-bsaviano:
> How about adding the command to the Explorer view title menu, perhaps with `go-to-file` as its icon?

You mean the default File Explorer, right? I will see if I can add the command there.

> Mapped .inc documents are not being hidden. To reproduce this, be connected to a namespace such as USER, set toggles to show system documents and to hide mapped ones. Observe that some %-packages still get listed. Drill into one and discover that you are being offered .inc documents despite them being mapped here from another database.

This sounds like a bug in the `%Library.RoutineMgr_StudioOpenDialog()` query. Please report it to the WRC.

> It would be nice if the three toggle buttons remembered their state between invocations of the selector, at least for the duration of my session, and perhaps persist per workspace.
> 
> It would also be nice if I could see the state of each. AFAIK the only way of doing this at the moment might be to create 6 of our own icons, 3 to represent "on" and 3 "off", then manipulate the `buttons` array. Perhaps the "on" icons would have a solid outline.
> 
> Alternatively I could have a go at a PR for [microsoft/vscode#185356](https://github.com/microsoft/vscode/issues/185356), or better still my suggestion at [microsoft/vscode#221397 (comment)](https://github.com/microsoft/vscode/issues/221397#issuecomment-2225357158) gets accepted.

I completely agree; this is why I opened that issue a while ago. I think your suggestion or the colored buttons would both be good solutions to this problem.

gjsjohnmurray:
> You mean the default File Explorer, right? I will see if I can add the command there.

Yes.

Another bit of feedback:

image

Could the title be reworded "Open a document in ..."?

isc-bsaviano:
@gjsjohnmurray I have that clumsy language to allow for the API caller to change the suffix but since we have no other callers now I can change that

isc-bsaviano:
@gjsjohnmurray I wasn't able to add the command inline using an icon, but it does appear in the `...` menu next to the other icons.

gjsjohnmurray:
Did you try adding `"group": "navigation"` to the new entry?

isc-bsaviano:
That worked! I tried `"group": "inline"` originally, which didn't work.

gjsjohnmurray:
> This sounds like a bug in the `%Library.RoutineMgr_StudioOpenDialog()` query. Please report it to the WRC.

Logged in WRC as 989244

gjsjohnmurray:
On a multiroot isfs workspace could the "Select workspace folder" selector default to the folder of the current active document rather thann the first folder?

isc-bsaviano:
The `showWorkspaceFolderPick()` doesn't allow you to set a default value. I suppose we could create a custom `QuickPick` with the default set, but that seems a little overkill to me.

gjsjohnmurray:
> I suppose we could create a custom QuickPick

I agree it's not worth the effort of doing that.

isc-bsaviano:
Thanks for all the feedback. I think this is ready to go. If/when we get toggle buttons or colored buttons for `QuickPick` I will add that to all places in ur extension that would benefit, including this command.

gjsjohnmurray:
> Logged in WRC as 989244

Confirmed by WRC, and Jira DP-433169 opened.

gjsjohnmurray (approved):
Great new feature!

gjsjohnmurray:
> > Logged in WRC as 989244
> 
> Confirmed by WRC, and Jira DP-433169 opened.

Fixed by MAK5922 (Do not display routines from IRISLIB/IRISSYS when 'mapped=0' parameter for StudioOpenDialog query) which may appear in 2024.3

PR #1399, opened 2024-07-12
Set file readonly if class is deployed or if server-side source control reports it not editable

### LINKED ISSUE #1397: Allow server-side file to readonly if server-side source control class for namespace returns that status
Some source control systems require code to be reserved/checked-out before it is changed.

Since ISFS was originally implemented VS Code's `FileSystemProvider` API has gained the ability for its `stat` method's [FileStat](https://code.visualstudio.com/api/references/vscode-api#FileStat) return value to include an optional `permissions` property of type [FilePermission](https://code.visualstudio.com/api/references/vscode-api#FilePermission). One of its bitmask values is `Readonly`.

I propose adding a boolean setting to make `stat` return `permissions` with the `Readonly` bit set or cleared depending on the value of the `Editable` boolean returned from the [GetStatus](https://docs.intersystems.com/irislatest/csp/documatic/%25CSP.Documatic.cls?LIBRARY=%25SYS&CLASSNAME=%25Studio.Extension.Base#GetStatus) method of the server-side source control class (if any) of the namespace.

The new setting would default to `false` so the new behaviour is opt-in and thus won't break existing environments.

It would be a resource-level setting so it can be configured server-side per namespace if desired.

gjsjohnmurray:
This PR closes #1397

In some circumstances the current VS Code (1.91.1) doesn't correctly switch an editor to be readonly. My PR https://github.com/microsoft/vscode/pull/221023 will resolve this. Meanwhile I am creating this PR as a Draft.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1310,12 +1310,14 @@
         "objectscript.serverSourceControl.disableOtherActionTriggers": {
           "description": "Prevent server-side source control 'other action' triggers from firing.",
           "type": "boolean",
-          "scope": "resource"
+          "scope": "resource",
+          "default": false
         },
         "objectscript.serverSourceControl.respectEditableStatus": {
           "markdownDescription": "Set `isfs` document readonly if GetStatus method of server-side source control class returns Editable = 0.",
           "type": "boolean",
-          "scope": "resource"
+          "scope": "resource",
+          "default": false
         },
         "objectscript.export": {
           "type": "object",
@@ -1379,11 +1381,13 @@
             },
             "atelier": {
               "description": "Export source code as Atelier did it, with packages as subfolders. This setting only affects classes, routines, include files and DFI files.",
-              "type": "boolean"
+              "type": "boolean",
+              "default": false
             },
             "generated": {
               "description": "Export generated source code files, such as INTs generated from classes.",
-              "type": "boolean"
+              "type": "boolean",
+              "default": false
             },
             "filter": {
               "markdownDescription": "SQL filter to limit what to export. The filter is applied to document names using the [LIKE predicate](https://irisdocs.intersystems.com/irislatest/csp/docbook/DocBook.UI.Page.cls?KEY=RSQL_like) (i.e. `Name LIKE '%filter%'`).",
@@ -1402,11 +1406,13 @@
             },
             "noStorage": {
               "description": "Strip the storage definition on export. Can be useful when working across multiple systems.",
-              "type": "boolean"
+              "type": "boolean",
+              "default": false
             },
             "dontExportIfNoChanges": {
               "description": "Do not rewrite the local file if the content is identical to what came from the server.",
-              "type": "boolean"
+              "type": "boolean",
+              "default": false
             },
             "maxConcurrentConnections": {
               "description": "Maximum number of concurrent export connections. (0 = unlimited)",
@@ -1414,7 +1420,8 @@
             },
             "mapped": {
               "description": "Export source code files mapped from a non-default database.",
-              "type": "boolean"
+              "type": "boolean",
+              "default": false
             }
           }
         },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1312,6 +1312,11 @@
           "type": "boolean",
           "scope": "resource"
         },
+        "objectscript.serverSourceControl.respectEditableStatus": {
```
isc-bsaviano, line 1315: Can we add an explicit default of `false`?

package.json
```diff
@@ -1312,6 +1312,11 @@
...
+          "description": "Set ISFS document readonly if GetStatus method of server-side source control class returns Editable=0",
```
isc-bsaviano, line 1316:
```suggestion
          "markdownDescription": "Set `isfs` document readonly if GetStatus method of server-side source control class returns Editable = 0.",
```

src/providers/FileSystemProvider/FileSystemProvider.ts
```diff
@@ -184,13 +192,55 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
...
+      result = await entryPromise;
+    } else {
+      // otherwise return the promise
+      return entryPromise;
+    }
+
+    //
+    if (result instanceof File) {
+      const api = new AtelierAPI(uri);
+      const serverName = getServerName(uri);
```
isc-bsaviano, line 225:
```suggestion
      const serverName = isCSPFile(uri) ? uri.path : uri.path.slice(1).replace(/\//g, ".");
```
This removes an unneeded dependency

src/providers/FileSystemProvider/FileSystemProvider.ts
```diff
@@ -184,13 +192,55 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
...
+      if (serverName.slice(-4).toLowerCase() == ".cls") {
+        if (await isClassDeployed(serverName, api)) {
+          result.permissions |= vscode.FilePermission.Readonly;
```
isc-bsaviano, line 228: It's a shame the API doesn't have a way to customize the read-only message too

src/providers/FileSystemProvider/FileSystemProvider.ts
```diff
@@ -184,13 +192,55 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
...
+          return result;
+        }
+      }
+
+      // Does server-side source control report it as editable?
+      if (vscode.workspace.getConfiguration("objectscript.serverSourceControl", uri)?.get("respectEditableStatus")) {
+        const query = "select * from %Atelier_v1_Utils.Extension_GetStatus(?)";
+        const statusObj = await api.actionQuery(query, [serverName]);
+        const docStatus = statusObj.result.content.pop();
```
isc-bsaviano, line 237:
```suggestion
        const docStatus = statusObj.result?.content?.pop();
```
Can't hurt to be very defensive

src/providers/FileSystemProvider/FileSystemProvider.ts
gjsjohnmurray: Indeed

package.json
gjsjohnmurray: Done, and also for other boolean properties where it was previously implicit.

gjsjohnmurray:
Set as ready so a testable VSIX gets built.

isc-bsaviano:
Thanks for making those changes

isc-bsaviano:
Should I wait until that linked Pull Request is merged before I test/approve this?

gjsjohnmurray:
The issue the linked PR fixes is only likely to affect how this new feature behaves with Deltanji's server-side source control class, because of how when we cancel a checkout we reset the mtime of the document to that of the reinstated version.

Please go ahead and review it.

gjsjohnmurray:
I pushed a change to remove those defaults.

As for the settings problem, I haven't been able to reproduce this. The setting displays for me in Settings Editor, and when present in JSON it's not marked / tipped as having a problem.

isc-bsaviano (approved):
@gjsjohnmurray I installed the latest dev vsix and everything looks good. Thanks for this PR!

PR #1416, opened 2024-07-26
Improve reliability of updating status bar panels

gjsjohnmurray:
This PR fixes two issues affecting our panels in the status bar.

1. If a multiroot workspace has documents open from two servers, clicking between the tabs wasn't updating the connection details on the status bar.

2. If MAC or INT documents were open, the line reference panel on the status bar would sometimes be uexpectedly empty. One case was when the workspace initially reloaded and output was displayed on the Output tab of the Panel.


isc-bsaviano (approved):
Looks good, but I didn't test

PR #1419, opened 2024-07-31
Add CSPSHARE=1 to align add-in behaviour with Studio

gjsjohnmurray:
The Studio add-in that is integral to George James Software's Yuzinji tool misbehaves when invoked from VS Code.

This PR corrects the issue by adding CSPSHARE=1 to the URL with which an add-in is launched.

Based on commit https://github.com/intersystems-community/vscode-objectscript/pull/756/commits/f58416c6719b0f75bede71ed58240227bf8160f2 by @timleavitt which was part of PR #756 I think it's very likely Studio adds this parameter too, which would explain the problem which this PR fixes.

### DIFF
--- a/src/commands/serverActions.ts
+++ b/src/commands/serverActions.ts
@@ -261,6 +261,7 @@ export async function serverActions(): Promise<void> {
                 params += `&Project=${encodeURIComponent(project)}`;
               }
               params += `&CSPCHD=${token}`;
+              params += "&CSPSHARE=1";
               vscode.env.openExternal(vscode.Uri.parse(`${serverUrl}${addin.id}?${params}`));
             }
           }


gjsjohnmurray:
Their browser will only have a working cookie if they have already run a template/add-in of the same type during this browser session. By "same type" I mean built-in (/isc/studio/templates) vs user-supplied (/isc/studio/usertemplates). I think we should be trying to make the feeling of integration between different components better rather than worse.

gjsjohnmurray:
In the Yuzinji case the user will actually need to authenticate twice when running the add-in for the first time in their external browser session. This is because, for sound architectural reasons, the add-in starts as a /isc/studio/usertemplates page and then redirects (client-side) to a /csp/yuzinji app.

The add-in also currently uses in its /csp/yuzinji pages some %session.Data that gets created in its /isc/studio/usertemplates pages. We _could_ re-engineer that if necessary.

isc-bsaviano:
@gjsjohnmurray I'm going to consult with our security team to make sure this use of CSPSHARE doesn't present a significant risk before I officially review the change.

isc-bsaviano (approved):
@gjsjohnmurray Our security team confirmed that using `CSPSHARE` in this case is fine. They did recommend removing the `CSHCHD` parameter from the links we generate for the SMP and documatic since those won't work on newer IRIS versions by default. I might do that in a separate PR.

PR #1438, opened 2024-09-30
Add `Launch Lite Terminal` action to Explorer

gjsjohnmurray:
This PR gives users of the client-side editing paradigm another way of launching a Lite Terminal.

### DIFF
--- a/src/commands/webSocketTerminal.ts
+++ b/src/commands/webSocketTerminal.ts
@@ -624,7 +624,10 @@ function terminalConfigForUri(
   }
 
   return {
-    name: api.config.serverName && api.config.serverName != "" ? api.config.serverName : "iris",
+    name:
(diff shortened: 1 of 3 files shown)


src/commands/webSocketTerminal.ts
```diff
@@ -624,7 +624,10 @@ function terminalConfigForUri(
   }
 
   return {
-    name: api.config.serverName && api.config.serverName != "" ? api.config.serverName : "iris",
```
isc-bsaviano, line 627: Not sure I like this part of the change. The host:port/prefix string can be very long and likely will be cut off in the UI by default.

gjsjohnmurray: Fair point. I have reverted this.

isc-bsaviano (approved):

PR #1442, opened 2024-10-03
Use more granular symbols for class members

### LINKED ISSUE #1441: Outline view mixes Foreign Key and Trigger definitions together with Properties
I have these items in my class definition:
 `ForeignKey CustomerFK(Owner) References Bank.Customer();`
`Trigger AccountUpdate [ Event = UPDATE, Foreach = row/object ]`

Their types **are** listed correctly in the Outline view.
However, they both use the Property icon (wrench) and they are sorted together with other properties. Also, the hover text is confusing because it shows "(property)".

Screenshot 2024-10-01 at 11 50 43 AM
Screenshot 2024-10-01 at 11 50 17 AM

isc-bsaviano:
This PR fixes #1441. When this PR is approved, I will make a corresponding Language Server commit.

Here's what the old symbols looked like:
before

And here's the new ones:
after

### DIFF
--- a/src/providers/WorkspaceSymbolProvider.ts
+++ b/src/providers/WorkspaceSymbolProvider.ts
@@ -38,19 +38,27 @@ export class WorkspaceSymbolProvider implements vscode.WorkspaceSymbolProvider {
     for (const element of data.result.content) {
       const kind: vscode.SymbolKind = (() => {
         switch (element.Type) {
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (approved):
LGTM

isc-rsingh (approved):

Issue #1427, opened 2024-09-06
Proposed APIs don't work on VS Code 1.93.0 if the server-side folder URI contains query parameters


RyszardWithAZ:
Originally it stopped working for me in VSCode Insiders a few weeks ago, but I thought it was me, and I needed to find whether I messed up a setting.

But my stable VSCode just upgraded to 1.93.0 and Ctrl-P stopped working there as well. It's possible they renamed their Quick Open API or did something with the FileSearchProvider?

isc-bsaviano:
@RyszardWithAZ Are you sure you have a beta version installed? This is working for me on VS Code 1.93.0 with extension version 2.12.8-beta.1. The VS Code team has been doing work on the `FileSearchProvider` and `TextSearchProvider` in preparation for finalizing them, but that work hasn't broken the existing proposals.

RyszardWithAZ:
I am running 
intersystems-community.vscode-objectscript version 2.12.8-beta.2 activating with proposed APIs available.
I also tried downgrading this to an older version and same deal.
That's weird that it's still working for you. I've done nothing else except let VSCode upgrade. Hm. And it's weird that my Insiders stopped working a while ago, and TODAY stable stopped working after I saw it upgraded.

isc-bsaviano:
Does the server-side text search work?

isc-bsaviano:
@RyszardWithAZ I believe this is a VS Code bug that only appears when the folder being searched has query parameters in the URI. I've opened https://github.com/microsoft/vscode/issues/227836 to report it.

RyszardWithAZ:
Hmm, the server-side text search also doesn't work.  Although now my entire shared dev env just went down. It's a Friday at 5:15. Yay. As soon as it comes back up (probably Monday), I can test that by rewriting my URI to not include any query parameters.

isc-bsaviano:
This issue has been fixed in the latest Insiders and the fix will be available in VS Code 1.95.0.

isc-bsaviano:
1.95.0 has been released

PR #1444, opened 2024-10-04
Make Explorer Find widget work in 1.94 as long as proposed APIs are enabled (fix #1443)

### LINKED ISSUE #1443: Since VS Code 1.94 the Explorer tree search no longer works for an ISFS folder
Type: <b>Bug</b>

1. Use Server Manager to create a workspace with an isfs folder accessing a namespace on a server.
2. Expand the folder to see some routines or classes.
3. While focused on Explorer tree, use Ctrl+Alt+F to open the tree's Find widget.
4. Enter some text that appears in a filename already visible.

The tree contents disappear and you're told no matches were found.

This is caused by a 1.94 change that isn't back-compatible.

I will prepare a PR to fix this.

Extension version: 2.12.9-beta.2
VS Code version: Code 1.94.0 (d78a74bcdfad14d5d3b1b782f87255d802b57511, 2024-10-02T13:08:12.626Z)
OS version: Windows_NT x64 10.0.22631
Modes:

gjsjohnmurray:
This PR fixes #1443 which occurs on VS Code 1.94+

A separate issue is that the Explorer Find widget now depends on FileSearchProvider. I will raise this with the VS Code project.

### DIFF
--- a/src/providers/FileSystemProvider/FileSearchProvider.ts
+++ b/src/providers/FileSystemProvider/FileSearchProvider.ts
@@ -18,6 +18,11 @@ export class FileSearchProvider implements vscode.FileSearchProvider {
   ): Promise<vscode.Uri[]> {
     let counter = 0;
     let pattern = query.pattern.charAt(0) == "/" ? query.pattern.slice(1) : query.pattern;
+
+    // Drop a leading **/ from the glob pattern if it exists (added by Find widget of Explorer tree, which since 1.94 uses FileSearchProvider)
+    if (pattern.startsWith("**/")) {
+      pattern = pattern.slice(3);
+    }
     const params = new URLSearchParams(options.folder.query);
     const csp = params.has("csp") && ["", "1"].includes(params.get("csp"));
     if (params.has("project") && params.get("project").length) {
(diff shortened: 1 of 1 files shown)


gjsjohnmurray:
Explorer Find remains broken for folders that use queryparams. This is almost certainly the same as #1427.

gjsjohnmurray:
1.94's new dependency on FileSearchProvider has been raised at https://github.com/microsoft/vscode/issues/230483

src/providers/FileSystemProvider/FileSearchProvider.ts
```diff
@@ -44,13 +49,17 @@ export class FileSearchProvider implements vscode.FileSearchProvider {
...
     if (pattern.length) {
-      pattern = !csp ? query.pattern.replace(/\//g, ".") : query.pattern;
+      let escapeClause = "";
+      pattern = !csp ? pattern.replace(/\//g, ".") : pattern;
       if (pattern.includes("_") || pattern.includes("%")) {
         // Need to escape any % or _ characters
-        filter = `Name LIKE '%${pattern.replace(/(_|%|\\)/g, "\\$1")}%' ESCAPE '\\'`;
-      } else {
-        filter = `Name LIKE '%${pattern}%'`;
+        pattern = pattern.replace(/(_|%|\\)/g, "\\$1");
```
isc-bsaviano, line 56: Why not always do this replacement and provide the ESCAPE clause? I doubt it impacts performance significantly and it makes this code shorter:

```ts
filter = `Name LIKE '%${pattern.replace(/(_|%|\\)/g, "\\$1").replace(/\*/g, "%").replace(/\?/g, "_")}%' ESCAPE '\\'`
```

isc-bsaviano (approved):

isc-bsaviano:
This looks good, assuming there are no globstars outside the possible leading one.

PR #1446, opened 2024-10-04
Fix fuzzy match in Explorer tree Find on an ISFS folder in 1.94 (fix #1445)

### LINKED ISSUE #1445: Fuzzy match in Explorer tree Find on an ISFS folder not fixed by #1443
#1444 got Explorer Find working again for 1.94 but didn't deal with fuzzy match mode.

Related to https://github.com/microsoft/vscode/issues/230483#issuecomment-2393928617

I will prepare a PR.

gjsjohnmurray:
This PR fixes #1445.

Still investigating some match failures, which may be caused by a VS Code bug. Hence PR is initially Draft.

### DIFF
--- a/src/providers/FileSystemProvider/FileSearchProvider.ts
+++ b/src/providers/FileSystemProvider/FileSearchProvider.ts
@@ -22,8 +22,8 @@ export class FileSearchProvider implements vscode.FileSearchProvider {
     // Drop a leading **/ from the glob pattern if it exists. This gets added by Find widget of Explorer tree (non-fuzzy mode), which since 1.94 uses FileSearchProvider
     if (pattern.startsWith("**/")) {
       pattern = pattern.slice(3);
-    } else if (pattern.length && !pattern.includes("*") && options.useGlobalIgnoreFiles) {
-      // Heuristic to detect a fuzzy search from Explorer's Find widget (see https://github.com/microsoft/vscode/issues/230483#issuecomment-2393928617)
+    } else if (pattern.length) {
+      // Do a fuzzy search
       pattern = "*" + pattern.split("").join("*") + "*";
     }
     const params = new URLSearchParams(options.folder.query);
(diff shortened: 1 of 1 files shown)


src/providers/FileSystemProvider/FileSearchProvider.ts
```diff
@@ -19,9 +19,12 @@ export class FileSearchProvider implements vscode.FileSearchProvider {
     let counter = 0;
     let pattern = query.pattern.charAt(0) == "/" ? query.pattern.slice(1) : query.pattern;
 
-    // Drop a leading **/ from the glob pattern if it exists (added by Find widget of Explorer tree, which since 1.94 uses FileSearchProvider)
+    // Drop a leading **/ from the glob pattern if it exists. This gets added by Find widget of Explorer tree (non-fuzzy mode), which since 1.94 uses FileSearchProvider
     if (pattern.startsWith("**/")) {
       pattern = pattern.slice(3);
+    } else if (pattern.length && !pattern.includes("*") && options.useGlobalIgnoreFiles) {
```
isc-bsaviano, line 25: I don't think this check is necessary. I think every search using this provider is supposed to be fuzzy. This is the documentation for `query.pattern`:

> The search pattern to match against file paths. To be correctly interpreted by Quick Open, this is interpreted in a relaxed way. The picker will apply its own highlighting and scoring on the results.
> 
> Tips for matching in Quick Open: With the pattern, the picker will use the file name and file paths to score each entry. The score will determine the ordering and filtering. The scoring prioritizes prefix and substring matching. Then, it checks and it checks whether the pattern's letters appear in the same order as in the target (file name and path). If a file does not match at all using our criteria, it will be omitted from Quick Open.

gjsjohnmurray:
The VS Code issue that can cause our fuzzy matches in the 1.94 Explorer to show incomplete results is noted at https://github.com/microsoft/vscode/pull/230486#issuecomment-2394516482

It will be fixed in Monday's Insiders and may get put into a 1.94 recovery. So I'm setting this PR as ready for review.

isc-bsaviano (approved):

PR #1467, opened 2025-01-23
Change some pickers from "workspace folders" to "server connections"

### LINKED ISSUE #1461: Lite terminal - default workspace folder
Hello,

Is there any option how to set default workspace folder to do not pick every time when I want to open lite terminal in VS code?

image

isc-bsaviano:
This PR fixes #1461. I added a new utility method for picking a server connection from the current workspace. It de-duplicates connections and has optional minimum server version checking. I used it for Lite Terminal and picking a process to attach to for debugging.

Screenshot 2025-01-23 at 9 12 22 AM

### DIFF
--- a/src/commands/newFile.ts
+++ b/src/commands/newFile.ts
@@ -247,7 +247,7 @@ export async function newFile(type: NewFileType): Promise<void> {
       wsFolder = vscode.workspace.workspaceFolders[0];
     } else {
       wsFolder = await vscode.window.showWorkspaceFolderPick({
-        placeHolder: "Pick the workspace folder to create the file in.",
+        placeHolder: "Pick the workspace folder where you want to create the file",
       });
     }
     if (!wsFolder) {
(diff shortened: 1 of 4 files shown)


isc-rsingh (dismissed):

gjsjohnmurray (changes requested):
Nice. Just one observation from my testing so far. When I have a workspace with at least 2 folders with distinct connections to 2023.2+ servers, and I invoke Lite Terminal from the dropdown on Panel's Terminal view, I correctly get the new picker. But if I change my mind and dismiss it with Esc I get an error notification on the bottom right saying "Lite Terminal requires an active server connection to InterSystems IRIS version 2023.2 or above.". It'd be nicer to suppress that in this case, but still present it if there were no eligible connections to offer the user. I guess that'd require `getWsServerConnection` to give different responses in those two cases rather than undefined for both.

PR #1471, opened 2025-01-31
Support Server Manager being able to handle `objectscript.conn.docker-compose` type connections

gjsjohnmurray:
This PR enables an incoming Server Manager enhancement to implement https://github.com/intersystems-community/intersystems-servermanager/issues/187

The VSIX built from this PR should work OK with Server Manager VSIXes that don't contain the relevant change, and vice versa. However the new features will only become available when both extensions have been updated.

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -110,7 +110,13 @@ export class AtelierAPI {
         if (schemas.includes(wsOrFile.scheme)) {
           workspaceFolderName = wsOrFile.authority;
           const parts = workspaceFolderName.split(":");
-          if (parts.length === 2 && config("intersystems.servers").has(parts[0].toLowerCase())) {
+          if (
+            parts.length === 2 &&
+            (config("intersystems.servers").has(parts[0].toLowerCase()) ||
+              vscode.workspace.workspaceFolders.find(
+                (ws) => ws.uri.scheme === "file" && ws.name.toLowerCase() === parts[0].toLowerCase()
+              ))
+          ) {
             workspaceFolderName = parts[0];
             namespace = parts[1];
           } else {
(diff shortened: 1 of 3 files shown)


src/api/index.ts
```diff
@@ -110,7 +110,13 @@ export class AtelierAPI {
         if (schemas.includes(wsOrFile.scheme)) {
           workspaceFolderName = wsOrFile.authority;
           const parts = workspaceFolderName.split(":");
-          if (parts.length === 2 && config("intersystems.servers").has(parts[0].toLowerCase())) {
+          if (
+            parts.length === 2 &&
+            (config("intersystems.servers").has(parts[0].toLowerCase()) ||
+              vscode.workspace.workspaceFolders.find(
+                (ws) => ws.uri.scheme === "file" && ws.name.toLowerCase() === parts[0].toLowerCase()
```
isc-bsaviano, line 117: Is it possible that someone would want to do this for non-`file` workspace folders?

src/extension.ts
```diff
@@ -1588,6 +1604,66 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
...
+        : vscode.workspace.getConfiguration(`intersystems.servers.${serverName.toLowerCase()}`, uri).get("password"),
+    namespace: ns,
+    apiVersion: active ? apiVersion : undefined,
+    serverVersion: active ? serverVersion : undefined,
+  };
+}
+
+// An async variant capable of resolving docker port number.
+// It is exported as one of our API functions but is also used internally.
+async function asyncServerForUri(uri: vscode.Uri): Promise<any> {
```
isc-bsaviano, line 1649: What does this mean for Language Server? Does it have to change to support these `docker compose` connections?

src/extension.ts
```diff
@@ -731,15 +780,20 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
   vscode.workspace.workspaceFolders?.map((workspaceFolder) => {
     const uri = workspaceFolder.uri;
     const { configName } = connectionTarget(uri);
-    const serverName = notIsfs(uri) ? config("conn", configName).server : configName;
+    const conn = config("conn", configName);
```
isc-bsaviano, line 783: Can we use the standard `vscode.workspace.getConfiguration()` API instead of this old custom one? It would be nice to phase that out in the future.

src/extension.ts
gjsjohnmurray: I don't think so. It will continue to ask the ObjectScript extension for the connection details (host, port, pathPrefix etc) as usual. This extension does the magic.

src/extension.ts
gjsjohnmurray: At this stage I'd prefer not to make unnecessary changes. The `config(...)` function does a number of things that mean I don't think `vscode.workspace.getConfiguration()` is a drop-in replacement.

isc-rsingh (approved):

PR #1477, opened 2025-02-05
Permit `Modify Server-side Workspace Folder...` before connection becomes active

### LINKED ISSUE #1476: `Modify Server-side Workspace Folder...` option missing from context menu of Explorer root folder until a server-side document has been opened
1. Use Server Manager to create a workspace accessing a namespace on a server.
2. Try to use Explorer context menu to invoke the prompt sequence for modifying the settings defaulted for the folder.

🐛 the option is missing until you have first opened a file from the server.

I will provide a PR.

gjsjohnmurray:
This PR fixes #1476

### DIFF
--- a/package.json
+++ b/package.json
@@ -606,7 +606,7 @@
         },
         {
           "command": "vscode-objectscript.modifyWsFolder",
-          "when": "explorerResourceIsRoot && !listMultiSelection",
+          "when": "(!resourceScheme || resourceScheme =~ /^isfs(-readonly)?$/) && explorerResourceIsRoot && !listMultiSelection",
           "group": "objectscript_modify@3"
         },
         {
--- a/package.json
+++ b/package.json
@@ -297,7 +297,7 @@
         },
         {
           "command": "vscode-objectscript.modifyWsFolder",
-          "when": "vscode-objectscript.connectActive && workspaceFolderCount != 0"
+          "when": "workspaceFolderCount != 0"
         },
         {
           "command": "vscode-objectscript.openErrorLocation",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -606,7 +606,7 @@
         },
         {
           "command": "vscode-objectscript.modifyWsFolder",
-          "when": "vscode-objectscript.connectActive && resourceScheme =~ /^isfs(-readonly)?$/ && explorerResourceIsRoot && !listMultiSelection",
```
isc-bsaviano, line 609: Removing the `vscode-objectscript.connectActive` check is fine, but why remove the `resourceScheme` one?

gjsjohnmurray: Because `resourceScheme` is undefined if you haven't opened a file yet. And in any case the file you've got open might not come from the root whose settings you're trying to alter.

isc-bsaviano: That sounds really strange to me. I just opened my `.code-workspace` file with two `isfs` folders. No files have been opened besides the "welcome" webview that opens automatically. The `Modify server-side workspace folder...` context menu option is present. Are those not valid reproduction steps?

gjsjohnmurray: Try starting with no workspace, then add a folder from Server Manager.

isc-bsaviano: Ok, now I can see it. This seems like a bug in VS Code though. The `resource*` context keys should be set to the Uri of the explorer tree item when you open the context menu. That works for saved workspaces upon open, but doesn't for untitled workspaces unless you open the folder that was just added.

gjsjohnmurray: It seems that the `resource*` context keys for the root folder are missing when it's a single-root workspace but are present when it's multi-root.

gjsjohnmurray: I plan to report it. Meanwhile I've tweaked my change.

isc-bsaviano (approved):

Issue #1263, opened 2023-11-01
Make a webapp ISFS folder fall back to the folder-specific settings of its namespace


gjsjohnmurray:
When using the server-side editing paradigm namespace folders can hold folder-specific settings (see doc at https://docs.intersystems.com/components/csp/docbook/DocBook.UI.Page.cls?KEY=GVSCO_ssworkflow#GVSCO_ssworkflow_config_folderspec).

When building a webapp using the technology formerly known as CSP it is common for a developer to need to edit code (e.g. classes) in the namespace associated with the webapp, as well as the webapp files (e.g. JS, CSS, and potentially also CSP / CSR). To do this a multi-root workspace can be created, e.g.

- isfs://iris:user/
- isfs://iris:user/csp/user?csp

There are circumstances (e.g. the Deltanji source control extension from George James Software) in which both folders should have the same folder-level settings. This is unsurprising since conceptually both are served by the same namespace.

Currently it is necessary to maintain two copies of the settings. On a default Windows install of IRIS the files would be:

- c:\InterSystems\IRIS\\.vscode\USER\settings.json
- c:\InterSystems\IRIS\csp\user\\.vscode\settings.json

On Linux and macOS the developer can probably create the latter folder to symlink to the former one, but this is harder (perhaps impossible) to achieve on Windows.

Also, in cases where a namespace hosts multiple webapps (e.g. `/csp/healthshare/namespace` and `/csp/healthshare/namespace/fhirconfig` and `/csp/healthshare/namespace/services`) it becomes necessary to set up and maintan multiple symlinks.

I propose we make the extension smart enough to spot when the uri `isfs://iris:user/csp/user/.vscode/settings.json?csp` returns FileNotFound, and in that case redirect to `isfs://iris:%SYS/_vscode/USER/settings.json?csp`.

Any site that needs different folder-level settings for a webapp compared to those they need for the associated namespace would manually create, say, `c:\InterSystems\IRIS\csp\user\.vscode\settings.json` as empty JSON (i.e. `{}`), and would thereafter be able to edit its contents using VS Code's Settings Editor.

gjsjohnmurray:
Closed by #1479

PR #1479, opened 2025-02-12
Make a webapp ISFS folder fall back to the folder-specific settings of its namespace

gjsjohnmurray:
This PR implements #1263

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -101,6 +101,7 @@ import {
   cspApps,
   otherDocExts,
   getWsServerConnection,
+  addWsServerRootFolderData,
 } from "./utils";
 import { ObjectScriptDiagnosticProvider } from "./providers/ObjectScriptDiagnosticProvider";
 import { DocumentLinkProvider } from "./providers/DocumentLinkProvider";
@@ -802,6 +803,9 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
       continue;
     }
   }
+  for await (const workspaceFolder of vscode.workspace.workspaceFolders) {
+    await addWsServerRootFolderData(workspaceFolder.uri);
+  }
 
   xmlContentProvider = new XmlContentProvider();
 
@@ -1320,6 +1324,9 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
         const uri = oneToCheck[1];
         const serverName = notIsfs(uri) ? config("conn", configName).server : configName;
         await resolveConnectionSpec(serverName);
+      }
+      for await (const workspaceFolder of added) {
+        await addWsServerRootFolderData(workspaceFolder.uri);
       }
     }),
     vscode.workspace.onDidChangeConfiguration(async ({ affectsConfiguration }) => {
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -617,6 +617,32 @@ export async function shellWithDocker(): Promise<vscode.Terminal> {
   return terminal;
 }
 
+interface WSServerRootFolderData {
+  redirectDotvscode: boolean;
+}
+
+const wsServerRootFolders = new Map<string, WSServerRootFolderData>();
+
+/**
+ * Add uri to the wsServerRootFolders map if eligible
+ */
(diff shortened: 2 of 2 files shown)


gjsjohnmurray:
Yes, that seems reasonable to me. If a user **does** want to create webapp-specific settings they'd only need to create a .vscode/settings.json stub file and reload/restart VS Code.

isc-bsaviano:
That sounds fine to me, just wanted to confirm that it was what you intended. I will review soon.

src/extension.ts
```diff
@@ -802,6 +803,9 @@ export async function activate(context: vscode.ExtensionContext): Promise<any> {
       continue;
     }
   }
+  for await (const workspaceFolder of vscode.workspace.workspaceFolders) {
```
isc-bsaviano, line 806: Do we have to `await` and delay activation? Also, this needs to handle the case where `workspaceFolders` is `undefined`.

src/utils/index.ts
```diff
@@ -617,6 +617,32 @@ export async function shellWithDocker(): Promise<vscode.Terminal> {
...
+  if (!schemas.includes(uri.scheme)) {
+    return;
+  }
+  const value: WSServerRootFolderData = {
+    redirectDotvscode: true,
+  };
+  if (isCSPFile(uri)) {
+    // A CSP-type root folder that already has a .vscode/settings.json file must not redirect .vscode/* references
+    const api = new AtelierAPI(uri);
+    api.getDoc(`${uri.path}/.vscode/settings.json`).then(() => {
```
isc-bsaviano, line 639: This should use `headDoc` instead, and needs a `catch`.

src/utils/index.ts
```diff
@@ -617,6 +617,32 @@ export async function shellWithDocker(): Promise<vscode.Terminal> {
...
+ * Add uri to the wsServerRootFolders map if eligible
+ */
+export async function addWsServerRootFolderData(uri: vscode.Uri): Promise<void> {
+  if (!schemas.includes(uri.scheme)) {
+    return;
+  }
+  const value: WSServerRootFolderData = {
+    redirectDotvscode: true,
+  };
+  if (isCSPFile(uri)) {
```
isc-bsaviano, line 636: You should add a check for `!["","/"].includes(uri.path)` because if that's true you have to redirect to the namespace version.

isc-bsaviano (changes requested):

src/extension.ts
gjsjohnmurray: Good catch about the possibility of it being undefined. Re await, I'd like to be sure that we've gathered the data before the FileSystemProvider starts being used. I'd expect that not to happen before activate has returned. So I'd prefer to leave the await in.

src/utils/index.ts
gjsjohnmurray: Done

src/utils/index.ts
gjsjohnmurray: I'm not clear what you mean.

isc-bsaviano: Sorry, I meant that if the root folder is showing all web apps and not a specific web app, then you need to redirect to the namespace settings.json anyways and there's no need to ping the server.

gjsjohnmurray: OK, I think I undertand.

src/utils/index.ts
```diff
@@ -617,6 +617,35 @@ export async function shellWithDocker(): Promise<vscode.Terminal> {
...
+    return;
+  }
+  const value: WSServerRootFolderData = {
+    redirectDotvscode: true,
+  };
+  if (isCSPFile(uri) && !["", "/"].includes(uri.path)) {
+    // A CSP-type root folder for a specific webapp that already has a .vscode/settings.json file must not redirect .vscode/* references
+    const api = new AtelierAPI(uri);
+    api
+      .headDoc(`${uri.path}/.vscode/settings.json`)
```
isc-bsaviano, line 640: Does this need to handle the case when `uri.path` ends with a `/`? Otherwise looks good.

gjsjohnmurray: Fixed

isc-bsaviano (approved):

PR #1484, opened 2025-02-13
Suffix ISFS folder name with web app path

### LINKED ISSUE #1483: Duplicate-named ISFS web files folders get added
1. Have a namespace that hosts more than one web application.
2. Use the `ObjectScript: Add Server Namespace to Workspace...` command to add a folder for one of the web applications.
3. Repeat previous step to add a folder for another web application using the same namespace.

:bug: both folders are named `server:NAMESPACE web files`

gjsjohnmurray:
This PR fixes #1483

### DIFF
--- a/src/commands/addServerNamespaceToWorkspace.ts
+++ b/src/commands/addServerNamespaceToWorkspace.ts
@@ -146,7 +146,7 @@ export async function addServerNamespaceToWorkspace(resource?: vscode.Uri): Prom
   const params = new URLSearchParams(uri.query);
   const project = params.get("project");
   const csp = params.has("csp");
-  const name = `${project ? `${project} - ` : ""}${serverName}:${namespace}${csp ? " web files" : ""}${
+  const name = `${project ? `${project} - ` : ""}${serverName}:${namespace}${!csp ? "" : ["", "/"].includes(uri.path) ? " web files" : ` web files (${uri.path})`}${
     scheme == FILESYSTEM_READONLY_SCHEMA && !project ? " (read-only)" : ""
   }`;
   // Append it to the workspace


src/commands/addServerNamespaceToWorkspace.ts
```diff
@@ -146,7 +146,7 @@ export async function addServerNamespaceToWorkspace(resource?: vscode.Uri): Prom
   const params = new URLSearchParams(uri.query);
   const project = params.get("project");
   const csp = params.has("csp");
-  const name = `${project ? `${project} - ` : ""}${serverName}:${namespace}${csp ? " web files" : ""}${
+  const name = `${project ? `${project} - ` : ""}${serverName}:${namespace}${!csp ? "" : ["", "/"].includes(uri.path) ? " web files" : ` web files (${uri.path})`}${
```
isc-bsaviano, line 149: Can we make the names shorter? How about `server:NS (Web)` and `server (app/path)`? We should change the project one to `server:NS (project)` as well to match.

isc-bsaviano (changes requested):

src/commands/addServerNamespaceToWorkspace.ts
gjsjohnmurray: Like this?

image

Seems OK as long as I don't have a project named "Web"

isc-bsaviano: I like that better, but I think we can remove the namespace when we're tied to a specific web app. IMO that extra info isn't worth the space in that case. I see what you mean with the potential web and project conflict. Do you want to leave the "web files" suffix then? I doubt that conflict will come up in reality but it could be confusing for a user if it does.

gjsjohnmurray: How about this?

image

isc-bsaviano: Looks good to me!

isc-bsaviano (approved):

src/commands/addServerNamespaceToWorkspace.ts
gjsjohnmurray: I'm having second thoughts about the change we've made to how a project folder gets named. Some people might not like how the project name is demoted. Here's the old way:

image

isc-bsaviano: Oh, I forgot that the project used to come first! I agree, let's put that back.

isc-bsaviano (approved):

PR #1485, opened 2025-02-14
Extend `objectscript.conn.docker-compose` settings object to handle superserver port identification

### LINKED ISSUE #1243: Extend `objectscript.conn.docker-compose` settings object to handle superserver port identification
The optional `objectscript.conn.docker-compose` object allows a VS Code workspace to connect to a local IRIS container instance for which the port serving the /api/atelier API outside the container is assigned dynamically at container startup rather than being hardcoded in the docker-compose.yml

The [vscode-iris-jupyter-server](https://github.com/intersystems-community/vscode-iris-jupyter-server) extension uses the Node Native API, so the connection has to be to the superserver port.

Let's extend this mechanism so it can look up the assigned superserver port and provide it in the connection object. Server Manager has already acquired the ability for superserver port (and host) to be defined (see https://github.com/intersystems-community/intersystems-servermanager/pull/200) but this doesn't help in the dynamic situation.

gjsjohnmurray:
This PR closes #1243

It will make the recently added `asyncServerForUri` API function resolve a Docker container connection's local superserver port into `server.superserverPort`, allowing future development in other extensions such as Server Manager to identify the superserver port of container-type connections as well as for ordinary ones.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1289,7 +1289,11 @@
                   "type": "string"
                 },
                 "internalPort": {
-                  "description": "Target port inside the service in docker-compose.",
+                  "description": "Target webserver port inside the service in docker-compose.",
+                  "type": "number"
+                },
+                "internalSuperserverPort": {
+                  "description": "Target superserver port inside the service in docker-compose.",
                   "type": "number"
                 },
                 "file": {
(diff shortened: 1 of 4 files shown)


isc-bsaviano (approved):
Looks good but I don't use this feature so have not tested

PR #1493, opened 2025-02-28
Implement custom workspace folder picker

isc-bsaviano:
- More info than built-in `showWorkspaceFolderPick()`.
- Allows for filtering invalid folders before prompting the user.
- Will not prompt the user if there is only one folder to pick from.
- Differentiates between user cancel and nothing to pick from.
Screenshot 2025-02-28 at 9 59 53 AM

### DIFF
--- a/src/commands/addServerNamespaceToWorkspace.ts
+++ b/src/commands/addServerNamespaceToWorkspace.ts
@@ -9,7 +9,7 @@ import {
   filesystemSchemas,
   smExtensionId,
 } from "../extension";
-import { cspAppsForUri, handleError, notIsfs } from "../utils";
+import { cspAppsForUri, getWsFolder, handleError, notIsfs } from "../utils";
 import { pickProject } from "./project";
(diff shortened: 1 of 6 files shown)


gjsjohnmurray (commented):
Code looks good, but I haven't tested it.

Now that we are in full control of the quickpick we could add a Close button to the titlebar to give a mouse-only way of dismissing it, since we're retaining the use of `ignoreFocusOut: true` which means clicking outside it doesn't dismiss it. See https://github.com/microsoft/vscode/issues/99081 for an open issue proposing doing this automatically some day.

However I don't feel strongly that we should do the extra work in this PR, so if you'd prefer not to then let me know and I'll approve it as it stands.

isc-bsaviano:
Thanks for reviewing and linking that issue. I was planning on doing a subsequent "UI normalization" PR that normalizes some language and the use of `title` instead of `placeHolder` so I can add this for some of our `QuickPicks` as part of that PR. I am also ok with removing `ignoreFocusOut` if we think that's inaccessible.

PR #1496, opened 2025-03-05
Input UI normalization

isc-bsaviano:
* Use `title` instead of `placeHolder` for inputs that only have one prompt
* Normalize language

I deliberately didn't do anything about `ignoreFocusOut` because I wanted to get your feedback first. Should we remove this from our extensions so there's always a mouse option to dismiss an input? Once this PR is merged I will make similar changes to the other two extensions.


gjsjohnmurray (dismissed):
Changes look fine.

We're likely overusing `ignoreFocusOut`, but at least some of the instances could have good justification so I don't think we should do a blanket removal. The question to put is: Might the user want to check information elsewhere on their screen while deciding which option(s) to pick?

gjsjohnmurray (approved):

PR #1503, opened 2025-03-10
Implement `Show Plan` CodeLens

isc-bsaviano:
This PR adds a new `Show Plan` CodeLens. The CodeLens will appear above `%SQLQuery` Class Query definitions and Embedded SQL queries if the query can generate an execution plan, and a connection to IRIS 2024.1+ is active. When the CodeLens is clicked, the query text, select mode and imported packages are extracted from the document and sent to the server. The returned plan is then rendered as HTML in a webview. Only one Show Plan webview can be visible at any time. Include files are captured, and support for passing them to the server-side SQL plan code will come in a future IRIS release.
Screenshot 2025-03-10 at 8 23 41 AM

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -3,10 +3,11 @@
 **
 !dist/*.js
 !dist/*.txt
-!snippets/
-!images/
-!syntaxes/
-!webview/
+!snippets/*.json
+!images/*.svg
+!images/*.png
+!syntaxes/*.json
+!webview/*.js
 !CHANGELOG.md
 !LICENSE
 !README.md
--- a/src/commands/documaticPreviewPanel.ts
+++ b/src/commands/documaticPreviewPanel.ts
@@ -38,7 +38,7 @@ export class DocumaticPreviewPanel {
    */
   public static currentPanel: DocumaticPreviewPanel | undefined;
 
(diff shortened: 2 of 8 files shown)


gjsjohnmurray:
@isc-bsaviano shall I delay looking at this until you have resolved the issues flagged by CI? And then is it time to ship new releases of this an Server Manager?

isc-bsaviano:
I didn't notice the CI failures, I will take a look at those now. Yes, this is the last change that I had planned for 3.0.0. I am out of the office this Thursday, Friday, and the following Monday so I will open the release preparation PRs for both extensions after I return.

gjsjohnmurray:
FYI I shall be OOO after 3/20, returning 4/2.

isc-bsaviano:
Thanks for letting me know. March 18th and 19th are the only days we'd both be available in case we need a quick patch version. I am fine with releasing one of those two days, or waiting until 4/2. @isc-rsingh Any thoughts?

gjsjohnmurray (approved):

isc-rsingh:
How about just waiting until 4/2 to release @isc-bsaviano and @gjsjohnmurray? That's my preference.

isc-bsaviano:
4/2 is fine with me.

PR #1505, opened 2025-03-11
Improve Lite Terminal shell integration

isc-bsaviano:
* Set the `IsWindows` [property](https://code.visualstudio.com/docs/terminal/shell-integration#_vs-code-custom-sequences-osc-633-st) to `False`
* Re-order shell integration sequences (https://github.com/microsoft/vscode/issues/242897)

### DIFF
--- a/src/commands/webSocketTerminal.ts
+++ b/src/commands/webSocketTerminal.ts
@@ -220,6 +220,8 @@ class WebSocketTerminal implements vscode.Pseudoterminal {
       this._closeEmitter.fire();
       return;
     }
+    // Set terminal properties
+    this._hideCursorWrite("\x1b]633;P;IsWindows=False\x07");
     // Print the opening message
     this._hideCursorWrite(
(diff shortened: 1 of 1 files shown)


src/commands/webSocketTerminal.ts
```diff
@@ -220,6 +220,8 @@ class WebSocketTerminal implements vscode.Pseudoterminal {
       this._closeEmitter.fire();
       return;
     }
+    // Set terminal properties
+    this._hideCursorWrite("\x1b]633;P;IsWindows=False\x07");
```
Tyriar, line 224: This is redundant fwiw as false is the default

Tyriar: FYI I think we're going to keep `HasRichCommandDetection` regardless of whether we end up doing the API in https://github.com/microsoft/vscode/issues/242897 as it's a very valuable signal for the dev and users to have. You would want to just write that all the time

isc-bsaviano: Thanks for the comments @Tyriar! I must have missed the `HasRichCommandDetecton` PR, I will adopt that sequence. While I have you attention, is there any chance that https://github.com/microsoft/vscode/issues/179913 can be looked at again?

PR #1514, opened 2025-03-18
Add confirmation step for XML export

### LINKED ISSUE #1509: Confirmation dialog before XML export
I'd still like an export confirmation dialog that just shows the selected files, rather than just having to scroll up and down the list to see which ones have been ticked :) 

_Originally posted by @mcwebtree in https://github.com/intersystems-community/vscode-objectscript/discussions/1504#discussioncomment-12468448_

isc-bsaviano:
This PR fixes #1509. Here's a screenshot of the new prompt. Let me know what you think. If we can't agree on a prompt then I suggest we close this PR and the linked issue.
Screenshot 2025-03-18 at 7 50 53 AM

### DIFF
--- a/src/commands/export.ts
+++ b/src/commands/export.ts
@@ -343,6 +343,27 @@ export async function exportDocumentsToXMLFile(): Promise<void> {
     if (documents.length == 0) {
       return;
     }
+    // Prompt the user to confirm their choices
+    const confirmed = await new Promise<boolean>((resolve) => {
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (approved):
Looks OK to me. Maybe not ideal UI, but a simple solution that shouldn't raise objections.

isc-rsingh (approved):

PR #1524, opened 2025-04-03
Fix issue where `Undo` after a save deletes the file being edited

### LINKED ISSUE #1522: v3.0.0 Saves file on every action with file (?)
When i try to discard git changes of local .CLS file, all file changes disappear from Source Control tab but one second later objectscript extension starts autosaving this file, no changes are made after save but file becomes "changed" for Source Control tab with actually no changes at all but still displayed in Source Control tab. 
Image

When i try to discard changes again, it starts autosaving again, file disappears for one second and then appears again. And it seems like there is no way out of this "loop" (metaphorically). 

Using `git reset --hard` doesn't help when vs code is opened. Can i disable this somehow?

isc-bsaviano:
This PR fixes #1522. The issue is that we "recreate" the file when loading changes from the compiler, so Undo undoes this "creation" and deletes the file. I fixed the issue by going back to the `fs.writeFile()` API. This issue is pretty bad, so I'm considering bumping version as part of the PR and doing the 3.0.1 release immediately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [3.0.1] 04-Apr-2025
 - Fixes
   - Fix issue where `Undo` after a save deletes the file being edited (#1524)
+  - Fix endless save loop when one local workspace folder is a subfolder of another (#1525)
 
 ## [3.0.0] 02-Apr-2025
 - Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## [3.0.1] 04-Apr-2025
+- Fixes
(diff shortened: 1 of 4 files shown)


gjsjohnmurray (dismissed):
LGTM but not tested

gjsjohnmurray (dismissed):

isc-rsingh (approved):
same

gjsjohnmurray (approved):

mathewrimmingtoncds:
Hi, do we have a delivery timeframe for this fix? Our developers are reporting the loss of significant bodies of work.

isc-bsaviano:
@mathewrimmingtoncds Version 3.0.1 is out.

PR #1528, opened 2025-04-09
Fix generating of URL when debugging REST services

### LINKED ISSUE #1527: Debugging REST Service builds improper URL since PR1487
We updated to version 3.0.1 and found that we cannot debug REST services anymore. When we try to start a REST service debug session, we are getting an Type Error: Invalid URL.

With [PR #1487](https://github.com/intersystems-community/vscode-objectscript/pull/1487) the WebApp dropdown was removed which is fine, but I was able to trace the error to the following code that still tries to build the URL using the WebApp value that is undefined creating an invaild URL.

In [src/commands/restDebugPanel.ts](https://github.com/intersystems-community/vscode-objectscript/blob/master/src/commands/restDebugPanel.ts) at line 361:

                url: `${encodeURI(`${serverInfo}${message.webApp}${path}`)}?${urlParams.toString()}`,

The value message.webApp is still in the code that is building the URL and builds an invalid URI. I believe the logic should either remove that value all together or it should be made conditional if message.webApp has a text value.

This was very easily reproduced in the latest copy from the master branch in debugging mode.

Let me know if you need more information. 

Thanks,
Mike

isc-bsaviano:
This PR fixes #1527

### DIFF
--- a/src/commands/restDebugPanel.ts
+++ b/src/commands/restDebugPanel.ts
@@ -20,8 +20,6 @@ interface WebviewMessage {
   bodyType: string;
   /** Request body */
   bodyContent: string;
-  /** Selected web application */
-  webApp: string;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


mloutraris:
Thanks for the prompt response!

gjsjohnmurray (approved):

Issue #1523, opened 2025-04-03
VSCode consuming more than one license unit per connection


VictorMGordillo:
Hello there,

I am afraid I need to report an issue about the license consumption from VSCode. This is after a join call with ISC support from the WRC, in which we diagnosed together the problem, and they were also able to replicate the same way.

Apparently, the /api/atelier application in the system is consuming way too many licenses when VSCode is opened and a class is being editted.

This is leading to systems being locked due to license usage, because the license count goes way too high.

The replication steps are simple:

1. Open your IRIS system SMP, and go to System Operation -> WebSessions
2. Remove all sessions.
3. Open then VSCode with just oine server connection, one namespace, and open one class for editting.
4. You will observe then back in the SMP how the web sessions are growing abnormally

$zversion for this test was IRIS for UNIX (Ubuntu Server LTS for x86-64) 2024.1.2 (Build 398U) Thu Oct 3 2024 13:38:26 EDT, and the ISC extensions were updated to the latest versions.

VictorMGordillo:
WRC# [998183] is the WRC number just in case you needed it...

isc-bsaviano:
Hi @VictorMGordillo, I will try to reproduce this. Did it start happening before or after the upgrade to 3.0.0 and server manager 3.10.0?

VictorMGordillo:
I am afraid I do not have this information, we just realized today of this problem

isc-bsaviano:
I was not able to reproduce this using the latest versions of all extensions (`3.0.1-beta.1`, `3.10.0` and `2.7.2`). I opened a server-side workspace folder with two folders connected to the same server, I opened a single class, and saved it a bunch of times. I hovered over some elements in the class so the Language Server would send some requests. I also fully expanded the Servers tree view in the InterSystems view container so Server Manager would make some requests. At the end, I still see the same three web sessions, which is expected. All of this was done using a connection to a server with a `$ZV` of `IRIS for UNIX (Ubuntu Server LTS for ARM64 Containers) 2024.1.2 (Build 398U) Thu Oct 3 2024 14:29:04 EDT`

bourette:
This is happening to my system currently.  With 3.0.1-SNAPSHOT and 3.10.0, and 2.7.2

isc-bsaviano:
@VictorMGordillo We;ve tried to reproduce this internall and find a cuase but have not been able to.

Issue #695, opened 2021-06-18
isfs workspace expansion fails silently if server definition is missing


gjsjohnmurray:
Issue Type: <b>Bug</b>

1. Create an isfs workspace.
2. Edit the code-workspace JSON, changing the server name to one that doesn't exist in your settings.json
3. Close and reopen the workspace, then try to expand the folder. The expanded tree node has no descendants but no explanation is given.

Extension version: 1.0.13-beta.3
VS Code version: Code 1.57.1 (507ce72a4466fbb27b715c3722558bb15afa9f48, 2021-06-17T13:28:07.755Z)
OS version: Windows_NT x64 10.0.19043
Restricted Mode: No

RyszardWithAZ:
I'm getting the same result often (but not always!), except my workspace has the correct server name. When I click on the Refresh Explorer icon, I briefly see my subpackages and then they disappear again.

Usually after I restart VSCode, it starts working again for a little while, until it stops. Of course, today it's not even recovering after a restart, I'm very frustrated I can't reach any of my code on the server.

gjsjohnmurray:
@RyszardWithAZ if you install the pre-beta build from https://github.com/intersystems-community/vscode-objectscript/suites/3032303382/artifacts/68794426 does the workspace folder change color when this happens during a refresh?

Is your server running out of licenses?

RyszardWithAZ:
I'm on the latest beta build (vscode-objectscript-1.0.13-beta.3), the workspace folder doesn't seem to be changing color, and I have PLENTY of licenses.

I'm also seeing save/compile operations in my output when I haven't saved anything. It looks like VSCode is CONSTANTLY saving to IRIS, even when I haven't saved the class. I'm also getting plenty of "Failed to save file on the server" messages, even if I'm not doing anything!

gjsjohnmurray:
Have you enabled autosave in VS Code?

RyszardWithAZ:
My workspace autosave is set to Off (though my User-level autosave is set to afterDelay)... I think my workspace one should win since I'm in my IRIS workspace?

PR #1532, opened 2025-04-10
Reduce REST traffic for server-side folders

isc-bsaviano:
This PR was prompted by my investigation into #1523.

- Don't call `GET /api/atelier/` every tie we try to read the root folder.
- Only call the `OpenedDocument` UserAction once per workspace folder.

### DIFF
--- a/src/providers/FileSystemProvider/FileSystemProvider.ts
+++ b/src/providers/FileSystemProvider/FileSystemProvider.ts
@@ -234,6 +234,7 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
   }
 
   public async stat(uri: vscode.Uri): Promise<vscode.FileStat> {
+    if (!new AtelierAPI(uri).active) throw vscode.FileSystemError.Unavailable("Server connection is inactive");
     let entryPromise: Promise<Entry>;
     let result: Entry;
     const redirectedUri = redirectDotvscodeRoot(uri);
--- a/src/providers/FileSystemProvider/FileSystemProvider.ts
+++ b/src/providers/FileSystemProvider/FileSystemProvider.ts
@@ -2,7 +2,7 @@ import * as path from "path";
 import * as vscode from "vscode";
 import { isText } from "istextorbinary";
 import { AtelierAPI } from "../../api";
-import { fireOtherStudioAction, OtherStudioAction, StudioActions } from "../../commands/studio";
+import { fireOtherStudioAction, OtherStudioAction } from "../../commands/studio";
 import { isfsConfig, projectContentsFromUri, studioOpenDialogFromURI } from "../../utils/FileProviderUtil";
 import {
   classNameRegex,
@@ -234,6 +234,9 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
   }
 
   public async stat(uri: vscode.Uri): Promise<vscode.FileStat> {
+    if (uri.path.includes(".vscode/") && !uri.path.endsWith("/settings.json")) {
+      throw vscode.FileSystemError.NoPermissions("Only settings.json is allowed within the /.vscode directory");
+    }
     let entryPromise: Promise<Entry>;
     let result: Entry;
     const redirectedUri = redirectDotvscodeRoot(uri);
@@ -284,19 +287,14 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
   }
 
   public async readDirectory(uri: vscode.Uri): Promise<[string, vscode.FileType][]> {
-    uri = redirectDotvscodeRoot(uri);
+    if (uri.path.includes(".vscode/")) {
+      throw vscode.FileSystemError.NoPermissions("Cannot read the /.vscode directory");
+    }
     const parent = await this._lookupAsDirectory(uri);
     const api = new AtelierAPI(uri);
     if (!api.active) throw vscode.FileSystemError.Unavailable(uri);
     const { csp, project } = isfsConfig(uri);
     if (project) {
-      if (["", "/"].includes(uri.path)) {
-        // Technically a project is a "document", so tell the server that we're opening it
-        await new StudioActions().fireProjectUserAction(api, project, OtherStudioAction.OpenedDocument).catch(() => {
-          // Swallow error because showing it is more disruptive than using a potentially outdated project definition
-        });
-      }
-
       // Get all items in the project
       return projectContentsFromUri(uri).then((entries) =>
         entries.map((entry) => {
@@ -405,7 +403,9 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
   }
 
   public createDirectory(uri: vscode.Uri): void | Thenable<void> {
-    uri = redirectDotvscodeRoot(uri);
+    if (uri.path.includes(".vscode/")) {
+      throw vscode.FileSystemError.NoPermissions("Cannot create a subdirectory of the /.vscode directory");
+    }
     const basename = path.posix.basename(uri.path);
     const dirname = uri.with({ path: path.posix.dirname(uri.path) });
     return this._lookupAsDirectory(dirname).then((parent) => {
@@ -421,6 +421,9 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
   }
 
   public async readFile(uri: vscode.Uri): Promise<Uint8Array> {
+    if (uri.path.includes(".vscode/") && !uri.path.endsWith("/settings.json")) {
+      throw vscode.FileSystemError.NoPermissions("Only settings.json is allowed within the /.vscode directory");
+    }
     // Use _lookup() instead of _lookupAsFile() so we send
     // our cached mtime with the GET /doc request if we have it
     return this._lookup(uri, true).then((file: File) => {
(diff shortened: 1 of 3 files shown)


isc-rsingh (dismissed):

gjsjohnmurray (changes requested):
I disagree that the .vscode folder support should be restricted to only the settings.json file. There is already value in being able to have a server-folder-specific launch.json file for debugging, .code-snippets file(s) and (to a lesser extent) tasks.json. Some extensions use .vscode/extensions to store stuff intended for sharing.

For this reason the suggested description of the /_vscode webapp at https://docs.intersystems.com/components/csp/docbook/DocBook.UI.Page.cls?KEY=GVSCO_serverflow#GVSCO_serverflow_folderspec is "Storage for VS Code namespace-specific settings, snippets, debug configurations etc."

gjsjohnmurray:
> * Only call the `OpenedDocument` UserAction once per workspace folder.

Is this consistent with how Studio behaved?

isc-bsaviano:
> Is this consistent with how Studio behaved?

I believe so. I'm pretty sure Studio only "opens" the project when it loads or the project is switched. 

> I disagree that the .vscode folder support should be restricted to only the settings.json file. There is already value in being able to have a server-folder-specific launch.json file for debugging, .code-snippets file(s) and (to a lesser extent) tasks.json. Some extensions use .vscode/extensions to store stuff intended for sharing.

VS Code probes the file system for these files constantly, and they never exist. This leads to a ton of junk traffic that clogs up debug logs and can pose problems for small user-based licenses if they cause the number of concurrent connections to rise above 25. I don't think that supporting other file types has value, regardless of the traffic impact. I only support doing this for `settings.json` because I know Deltanji needs it.

gjsjohnmurray:
If there's a problem with concurrent connections doesn't this indicate we're not providing a session cookie properly in some cases?

What if the first time on a given connection the extension needs to check existence of .vscode/X.y it asks the server, then caches the name if it doesn't exist and thereafter uses the cache to avoid the REST call? If the user subsequently creates .vscode/X.y through VS Code and saves it to the server we'd remove its name from the cache.

We could even do this for settings.json as long as users understand that after one member of the team creates that file for the first time for a shared namespace the other team members would need to reload their VS Code window, or reconnect. Alternatively we could treat settings.json as a special case and always test its existence on the server rather than adding it to the non-existence cache.

isc-bsaviano:
> If there's a problem with concurrent connections doesn't this indicate we're not providing a session cookie properly in some cases?

We are properly using session cookies. The issue is that if a logged in user creates more than a certain number of concurrent connections, each of those connections will take out a license unit instead of them all sharing one unit. This can cause license usage to spike. Removing this traffic should help reduce the chances of that happening.

> What if the first time on a given connection the extension needs to check existence of .vscode/X.y it asks the server, then caches the name if it doesn't exist and thereafter uses the cache to avoid the REST call? If the user subsequently creates .vscode/X.y through VS Code and saves it to the server we'd remove its name from the cache.

I would rather us make a decision on what we are going to support. I think that users advanced enough to set up the /_vscode web app are going to expect the stored settings to behave like the rest of VS Code settings and update dynamically. Therefore, I think we should allow the constant pinging of `settings.json`. However, I don't think there's a good reason to support storing any other file. Debug configurations can be put in the `.code-workspace`, and tasks don't make sense in our server-side environment. I don't see anything [in the docs](https://code.visualstudio.com/docs/editing/userdefinedsnippets) about snippets being stored in the workspace.

gjsjohnmurray:
> I don't see anything [in the docs](https://code.visualstudio.com/docs/editing/userdefinedsnippets) about snippets being stored in the workspace.

Using the `Snippets: Configure Snippets...` command on a 2-namespace ISFS workspace:

image

Doc reference: https://code.visualstudio.com/docs/editing/userdefinedsnippets#_project-snippet-scope

gjsjohnmurray:
> Debug configurations can be put in the `.code-workspace`

So can settings, and like settings there are scenarios where a debug config tailored to a specific namespace can be valuable.

gjsjohnmurray:
As for license starvation, I ran this experiment, starting with no VS Code instances open:

1. Restart IRIS.
2. Log in as me on IRIS Portal and go to System Operation> License Usage > Usage by User
3. Observe 1 connection, 1 web session, 1 unit against my ID.
4. Open a fresh VS Code and use Server Manager to connect to the server as me (i.e. expand it to see the namespaces subfolder).
5. Before even expanding the Namespaces subfolder in Server Manager, in IRIS Portal refresh page and observe my ID having 3 connections, 3 web sessions, 1 unit.
6. Close VS Code, then refresh the Portal page. My counts are still 3, 3, 1
7. Repeat step 4, then step 5. My counts are now 5, 5, 1.
8. Another repeat will take them to 7, 7, 1 and so on until we reach 25, 25, 1.
9. The next attempt will give an "Unavailable" label instead of the "Namespaces" folder, because my IRIS license is a 10-user one so can't trade me up to using 27 units rather than 1.

Can we do anything to release connections when closing VS Code?

isc-bsaviano:
> Doc reference: https://code.visualstudio.com/docs/editing/userdefinedsnippets#_project-snippet-scope

Thanks for the correction.

> So can settings, and like settings there are scenarios where a debug config tailored to a specific namespace can be valuable.

I just disagree that supporting these other files is valuable. AFAIK there aren't any external tools that depend on files other than settings.json. IMO it's strange architecturally that we persistent VS Code client configuration on the server. The "server-namespace" storage strategy doesn't neatly map to the "workspace folder" VS Code setting concept. I would prefer to not do this except for cases where it's needed, like settings.

> Can we do anything to release connections when closing VS Code?

I don't think this is the cause of that issue because it seems to be happening on open and without much user interaction. However, this may be something good for our extensions to do in the `deactivate()` function.

isc-bsaviano:
@gjsjohnmurray I removed the restriction on files within the `.vscode` folder.  I would like to revisit that, but it should be in a separate PR.

gjsjohnmurray:
> @gjsjohnmurray I removed the restriction on files within the `.vscode` folder. I would like to revisit that, but it should be in a separate PR.

@isc-bsaviano I'd like to understand why some of your original changes in FileSystemProvider.ts relating to the .vscode folder are still being retained in this PR.

src/providers/FileSystemProvider/FileSystemProvider.ts
```diff
@@ -405,7 +400,9 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
   }
 
   public createDirectory(uri: vscode.Uri): void | Thenable<void> {
-    uri = redirectDotvscodeRoot(uri);
+    if (uri.path.includes(".vscode/")) {
```
isc-bsaviano, line 403: This can probably be removed since we don't actually call to the server to create a directory.

src/providers/FileSystemProvider/FileSystemProvider.ts
```diff
@@ -284,19 +284,14 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
   }
 
   public async readDirectory(uri: vscode.Uri): Promise<[string, vscode.FileType][]> {
-    uri = redirectDotvscodeRoot(uri);
+    if (uri.path.includes(".vscode/")) {
```
isc-bsaviano, line 287: This prevents VS Code from sending a StudioOpenDialog query request on the "_vscode" folder, which won't return anything. I tested that VS Code still uses the server-side settings.json file if it exists with this check added.

src/providers/FileSystemProvider/FileSystemProvider.ts
```diff
@@ -699,6 +696,9 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
     if (vscode.workspace.getWorkspaceFolder(oldUri) != vscode.workspace.getWorkspaceFolder(newUri)) {
       throw vscode.FileSystemError.NoPermissions("Cannot rename a file across workspace folders");
     }
+    if (oldUri.path.includes(".vscode/") || newUri.path.includes(".vscode/")) {
```
isc-bsaviano, line 699: VS Code will never do this, so it's purely precautionary.

src/providers/FileSystemProvider/FileSystemProvider.ts
gjsjohnmurray: In that case I suggest it gets removed to avoid unnecessary changes.

src/providers/FileSystemProvider/FileSystemProvider.ts
gjsjohnmurray: Please either remove it or add an explanatory comment to the code.

src/providers/FileSystemProvider/FileSystemProvider.ts
```diff
@@ -864,11 +859,6 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
   // Fetch entry (a file or directory) from cache, else from server
   private async _lookup(uri: vscode.Uri, fillInPath?: boolean): Promise<Entry> {
     const api = new AtelierAPI(uri);
-    if (uri.path === "/") {
```
gjsjohnmurray, line 867: What's being removed here seems to relate to my fix for #695 from 4 years ago. IIRC that issue could arise if I create XYZ.code-workspace with a folder whose isfs uri references a server definition name that's in my settings.json, then I send you the workspace file to use but you don't have the server definition set up. Rather than failing silently this code raises a message.

Do the benefits of removing it in this PR outweigh those of retaining it?

isc-bsaviano: Yes, removing this line did allow that bug to reappear. I've found another way to fix it, see my latest commit.

gjsjohnmurray (approved):

PR #1538, opened 2025-04-25
Update text of all local files changed by compilation

### LINKED ISSUE #1537: missing changed storage info when compiling multiple classes
Type: <b>Bug</b>

when compiling a class which inherits from other classes, sometimes it is possible that compilation modifies storage of those other classes. However, these changes are not sent back from IRIS server to local files, thus a discrepancy occurs between local files and their server copies. This behaviour is leading to issues when using GIT (or other source control) and deployment of changes to other IRIS server(s) as compilation on those other servers may generate different storage changes to dependent classes.

Extension version: 3.0.1
VS Code version: Code 1.99.3 (17baf841131aa23349f217ca7c570c76ee87b957, 2025-04-15T23:18:46.076Z)
OS version: Windows_NT x64 10.0.26100
Modes:

isc-bsaviano:
This PR fixes #1537. Note that this PR only addresses the issue for client-side editing workflows. The server-side work is much more involved and therefore should be done separately.

### DIFF
--- a/src/commands/compile.ts
+++ b/src/commands/compile.ts
@@ -34,7 +34,7 @@ import {
 } from "../utils";
 import { StudioActions } from "./studio";
 import { NodeBase, PackageNode, RootNode } from "../explorer/nodes";
-import { updateIndexForDocument } from "../utils/documentIndex";
+import { getUrisForDocument, updateIndexForDocument } from "../utils/documentIndex";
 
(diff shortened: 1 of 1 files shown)


isc-rsingh (approved):

gjsjohnmurray (approved):
LGTM, but not tested

PR #1543, opened 2025-05-01
Add button to show Output channel if it's not visible for compile errors

### LINKED ISSUE #1542: When there is a compilation error, the button to show output is no longer there
Image
There used to be a button on this popup which would open the output so you could see the error. Now it's not there, it's a slightly annoying manual process to see the actual compilation error. As someone who likes the terminal minimised until I need it, it would be quite nice to have the button back if possible.

isc-bsaviano:
This PR fixes #1542


gjsjohnmurray (approved):
LGTM but not tried

PR #1544, opened 2025-05-02
Small tweaks

isc-bsaviano:
This PR arose during my investigation of #1537 for server-side editing.

- Fire file change events for related documents that were updated by compilation.
- Remove unneeded cache clearing before firing file change events and refresh of server-side class that's open in a low-code editor. VS Code's calling of `stat()` will cause the file to update in these cases.
- Remove unneeded `workspaceState` updates from FileSystemProvider. This is a client-side feature.
- Fire file create events when creating files via `writeFile()`or `rename()`.
- Don't allow `rename()` to overwrite a file that's marked read-only by source control or due to being deployed.
- Don't allow using client-side "import and compile" commands with server-side files.
- Hide some commands when the active text editor is dirty.

@gjsjohnmurray I've tested this a bunch but I would appreciate you playing with it for a little bit just to make sure it's good.

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,15 +105,15 @@
         },
         {
           "command": "vscode-objectscript.compile",
-          "when": "editorLangId =~ /^objectscript/ && vscode-objectscript.connectActive"
+          "when": "!(resourceScheme =~ /^isfs(-readonly)?$/) && editorLangId =~ /^objectscript/ && vscode-objectscript.connectActive && !activeEditorIsDirty"
         },
         {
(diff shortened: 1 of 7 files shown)


isc-bsaviano:
@gjsjohnmurray How does this look? I would like to get these last two PRs merged soon do I can do a 3.0.2 release next week.

gjsjohnmurray (approved):
I have used the dev VSIX for a couple of days (server-side paradigm only) and not seen any problems, so am approving this.

isc-bsaviano:
Thanks John!

PR #1555, opened 2025-05-16
Don't report error when deleting a local file that doesn't exist on the server

isc-bsaviano:
See comment in code for reasoning


isc-rsingh (approved):

gjsjohnmurray (approved):
Seems reasonable to me

PR #1563, opened 2025-05-23
An alternative for autoAdjustName

### LINKED ISSUE #1559: objectscript.autoAdjustName behavior changed on client-side editing
In version 3.0.0, the behavior of the aforementioned configuration was adjusted:

 - Change the default value of the `objectscript.autoAdjustName` setting to `false`. Now that we have an index of the workspace, we no longer require that a document's name match the file path for the extensions to find it. This setting only affects files that are copied or moved. New files will still have the Class or ROUTINE header generated upon file creation.

However, this change seems to have led to unexpected side effects in client-side editing:
Problem 1: When creating a blank file, the plugin may automatically generate a blank class definition.
Problem 2: When saving server code to disk using Ctrl+S in the Server Explorer, the plugin may unexpectedly modify CLASS package names or ROUTINE names.

Problem 1 may be acceptable, but Problem 2 is not, as it modifies existing code without explicit user consent. This behavior risks users unknowingly committing incorrect code if they fail to notice the altered package name.

Could you confirm if this is the intended behavior and provide some insight into the reasoning behind it?

hsyhhssyy:
This PR fixes #1559

image

We could turn autoAdjustName into a selectable setting. 
The default option, Create, would adjust names only when a file is created—matching the current behavior when the flag is set to false. 
Choosing Create, Copy, and Move would mimic the behavior of true. 
Finally, selecting Never would leave files on disk exactly as they are, which is the outcome I personally prefer.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "vscode-objectscript",
-  "version": "3.0.0-SNAPSHOT",
+  "version": "3.0.3-SNAPSHOT",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "vscode-objectscript",
-      "version": "3.0.0-SNAPSHOT",
+      "version": "3.0.3-SNAPSHOT",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1476,9 +1476,19 @@
           "default": false
         },
         "objectscript.autoAdjustName": {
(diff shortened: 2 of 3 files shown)


hsyhhssyy:
Might be a bug: when you select “Create, Copy and Move” (or set AutoAdjustName to yes as it works now) and drag-and-drop to move a file from one workspace folder to another, the file’s contents stay the same.

isc-bsaviano:
Hi @hsyhhssyy, I appreciate the PR but after reviewing this with another maintainer we've decided to reject it. We think it's desirable for the stubs to always be generated for new files. Since the stubs are not auto-saved, you can remove them using `Ctrl-Z` if they are unwanted. Thanks for your interest in this extension!

PR #1564, opened 2025-05-23
Fix detection of duplicate file change events

### LINKED ISSUE #1550: Export issues
Hi,
I am currently using v3.0.2-beta.12 to test the Git functionality and I have two problems with the export. Latest 2.x version seems to be fine. 

When I run "Export" in the Iris-Explorer:

- It re-compiles all the exported files.
Image
- It opens all files in status "changed". When I save them after the export, Git does not show any changes.
Image

br Matthias

isc-bsaviano:
This PR fixes #1550 and improves on the original fix from #1539.


isc-bsaviano:
I will wait to merge until the issue reporter lets me know that the fix works for them. They will be using it for a few days first.

PR #1573, opened 2025-06-04
Update the VS Code editor UI if a server-side document was changed during save

### LINKED ISSUE #1565: Server Side changes no longer cause document to reload
We're using a version of isc.codetidy that enforces consistent formatting in our server side source control hooks. We've found that since version 3.0.0 of the ObjectScript plugin the changes are no longer causing the modified document to appear in VSCode. The hooks are running correctly - if I create or update a document using the Atelier API directly, the content is returned with the correct formatting. Closing and reloading a document causes the updated formatting to appear. If I roll back to 2.12.10 of the plugin then the document is reloaded after saving. 

I've attached my example API calls made using Postman. 

[atelier-response.json](https://github.com/user-attachments/files/20448913/atelier-response.json)

[atelier-body.json](https://github.com/user-attachments/files/20448915/atelier-body.json)

$zv = IRIS for UNIX (Red Hat Enterprise Linux 9 for x86-64) 2024.1.1 (Build 347_0_24046U) Thu Apr 3 2025 00:25:59 EDT

isc-bsaviano:
This PR fixes #1565. Calling the revert command looks a little hacky, but the end result is the same as the version 2.x behavior.

### DIFF
--- a/src/providers/FileSystemProvider/FileSystemProvider.ts
+++ b/src/providers/FileSystemProvider/FileSystemProvider.ts
@@ -441,13 +441,15 @@ export class FileSystemProvider implements vscode.FileSystemProvider {
     }
     const api = new AtelierAPI(uri);
     let created = false;
+    let update = false;
+    const isCls = !csp && fileName.split(".").pop().toLowerCase() == "cls";
     // Use _lookup() instead of _lookupAsFile() so we send
     // our cached mtime with the GET /doc request if we have it
     return this._lookup(uri)
       .then(
         async (entry: File) => {
           // Check cases for which we should fail the write and leave the document dirty if changed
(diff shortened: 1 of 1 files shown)


gjsjohnmurray:
From a quick read of the code I'm not sure if it always uses revert for classes. Even if it does, I don't recall if 2.x worked that way too.

gjsjohnmurray (approved):
Once merged I will start using the beta and look out for any unexpected changes of behaviour during my everyday server-side editing.

isc-bsaviano:
Thanks @gjsjohnmurray. I plan on doing a 3.0.3 release by the end of the day next Monday so that leaves plenty of time for beta verification.

PR #1574, opened 2025-06-04
Don't show `Show Plan` CodeLenses inside string literals

isc-bsaviano:


gjsjohnmurray (approved):
Looks OK but not personally tested.

PR #1575, opened 2025-06-05
Only reload contents of client-side file after compilation if compilation changed the file

isc-bsaviano:
This PR modifies the post-compile reloading of client-side files to only occur if compilation changed the file's contents (for example, modifying the Storage). The current behavior always reloads the file's contents. This is rarely necessary and leads to a bunch of extra REST requests and file system events.

Created this as a draft because while it's an improvement architecturally, I'm not sure if it necessarily improves UX. The reloading will replace the contents of classes with the contents in the server "canonical" form. Even when not necessarily functionally, some users may want this enforced to avoid larger source control diffs when the class does actually need to be reloaded due to compilation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Change Log
 
+## [3.2.0] 06-Oct-2025
+Minimum VS Code version is now 1.104.0.
+- Enhancements
+  - Better error handling when expanding InterSystems and Projects Explorers (#1652)
+  - Allow opening low-code editors from read-only file systems (#1655)
+  - Add command for showing all class members, including inherited (#1656)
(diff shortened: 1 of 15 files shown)


isc-bsaviano:
Spoke to Raj about this and we decided that it needs more design work and then testing with users before it could be merged. I will leave it in Draft and keep updating the branch after releases. Here are some questions to resolve:

1. Should changes made due to server-side source control be ignored?
2. Should this be configurable?
3. For Storage-only changes (the vast majority, if not all changes), should we attempt to merge the new Storage into the existing class so the whole class doesn't need to be updated?
4. Does `objectscript.multilineMethodArgs` become a server-side-only setting?
5. Ideally this would be made to play nicely with a future Prettier-style formatting tool.

PR #1602, opened 2025-07-01
Check if `/_vscode` web app exists before redirecting `.vscode` files

isc-bsaviano:
This PR reduces the number of unnecessary for REST requests for server-side folders that cannot redirect `.vscode` files. To do this, I check the list of %SYS web apps to see if `/_vscode` is in it. I verified that VS Code behaved correct when the `/_vscode` web app was enabled or disabled. While working on this change, I discovered a few other issues that I've fixed:

- Cache the list of web apps and abstract document types for all server-namespaces in the workspace.
- Combined our `onDidChangeWorkspaceFolders` event handlers into a single function.
- Make some FileSystemProvider errors shorter (see #1587).

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -446,21 +446,6 @@ export async function checkConnection(
         serverVersion: info.result.content.version,
         healthshare: hasHS ? "yes" : "no",
       });
-
-    // Update CSP web app cache if required
-    const key = `${api.serverId}:${api.config.ns}`.toLowerCase();
-    if (!cspApps.has(key)) {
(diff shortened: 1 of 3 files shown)


isc-rsingh (approved):

gjsjohnmurray:
Doc for FileSystemError says (my added emphasis):

> A type that filesystem providers **should** use to signal errors.

I'd be inclined to play safe and toe the line here.

gjsjohnmurray (approved):

PR #1616, opened 2025-07-16
XML import/export issues

### LINKED ISSUE #1614: Importing XML files bypasses server side source control
I have an iris server with Deltanji server side source control. 

If I try to edit any files, the source control forces me to check out the files I wish to work on. This is as expected and desired. 

If I import an XML file with changes from another location, VS Code allows the import and deploys the files to the server without any intervention from the source control.  I can then open the files on the server in VSCode and see the changes to the files. This is NOT as expected or desired. 

If I then go to modify the files further in VSCode, I'm forced to check out as expected. 
The standard Deltanji actions then overwrite the imported changes with the source controlled version from the repository, (again expected and desired behaviour). 

The issue is that the import of XML does not invoke any source control actions so invalidates the integrity of the source controlled environment, and because those imported changes are rolled back when a file is later edited could result in dangerous code regression errors which become very hard to trace.

### LINKED ISSUE #1615: XML export output lines
When you export a file in Studio you get the following output: 

Image

as in after the "Export finished successfully" message you get the location and name of the exported file. 

I find this really useful, however VSCode skips that final line and I can't see anywhere that you can check where on earth you put your export file. 

I often just copy the file path to use as the import for another IDE session on another server which saves a lot of browsing time since, besides other issues, the last import/export file still doesn't seem to be cached/remembered by VSCode. 

Image

Can that last line be added back in please? 

Many thanks.
Mark

isc-bsaviano:
This PR fixes #1614 and fixes #1615.
- Skip importing documents that are marked read-only by server-side source control when importing from XML files.
- Output the path to the export file when exporting to an XML file.

### DIFF
--- a/src/commands/export.ts
+++ b/src/commands/export.ts
@@ -352,7 +352,6 @@ export async function exportDocumentsToXMLFile(): Promise<void> {
       quickPick.title = `Export the following ${documents.length > 1 ? `${documents.length} documents` : "document"}?`;
       quickPick.placeholder = "Click any item to confirm, or 'Escape' to cancel";
       quickPick.ignoreFocusOut = true;
-      quickPick.onDidChangeSelection((e) => outputChannel.appendLine(JSON.stringify(e)));
       quickPick.onDidAccept(() => {
         resolve(true);
         quickPick.hide();


isc-bsaviano:
Console output showing the changes:
Screenshot 2025-07-17 at 6 58 13 AM

Screenshot 2025-07-17 at 7 00 40 AM

gjsjohnmurray (approved):
LGTM and I'm happy to approve it. Up to you whether you prefer to wait for confirmation from @mcwebtree that he tested the dev build, or just go ahead with the merge anyway.

PR #1619, opened 2025-07-18
Fix breakage v3.0 caused to the InterSystems Testing Manager extension

### LINKED ISSUE #1618: v3.0 broke the InterSystems Testing Manager extension's operation when used with client-side editing paradigm
Ever since #1479 my InterSystems Testing Manager extension has been unable to run tests when the client-side editing paradigm is being used.

See https://github.com/intersystems-community/intersystems-testingmanager/issues/41

gjsjohnmurray:
This PR fixes #1618

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -605,25 +605,39 @@ export async function addWsServerRootFolderData(wsFolders: readonly vscode.Works
   if (!wsFolders?.length) return;
   return Promise.allSettled(
     wsFolders.map(async (wsFolder) => {
-      if (notIsfs(wsFolder.uri)) return;
       const api = new AtelierAPI(wsFolder.uri);
       if (!api.active) return;
       const value: WSServerRootFolderData = {
         redirectDotvscode: true,
         canRedirectDotvscode: true,
       };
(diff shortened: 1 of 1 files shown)


isc-bsaviano:
@gjsjohnmurray Can you explain why this is necessary? Why can't the testing manager generate the redirected .vscode URI itself?

gjsjohnmurray:
I'd prefer not to have to make the Testing Manager have to deal with /api/atelier file- and directory-type endpoints directly. But perhaps I can make it talk to isfs://server/_vscode/_NAMESPACE_/foo.bar?csp URLs directly instead of using isfs://server/.vscode/foo.bar and relying on the the redirection (which is what's no longer happening for client-type workspaces since #1479). I'll try this out in Testing Manager.

isc-bsaviano:
Yeah, I can see why it would be easier to use `vscode.workspace.fs` on `isfs` URIs. I think using `/_vscode` directly should work though, assuming there really is a `/_vscode` web app on that server.

gjsjohnmurray:
I worked around it in the other extension so I'm closing this unmerged.

PR #1622, opened 2025-07-18
Fix resolving of server connection for files opened from InterSystems Explorer or XML preview

isc-bsaviano:
Since `objectscript(xml)://` URIs are not in a workspace folder, resolving server connection information for them didn't always work. This lead to issues like a lack of intellisense in those files or the server connection for the current workspace being forced inactive. This also caused Language Server to prompt for a password when not necessary because this extension would not return the password when it was defined in plain text.


isc-bsaviano:
@gjsjohnmurray @isc-rsingh I would appreciate a review on this today so we can proceed with the release.

gjsjohnmurray (approved):
Looks OK but not personally tested.

PR #1634, opened 2025-08-20
Fix CSP files not being included in server side search results

psteiwer:
Bug with the expected value of "projects". This was leading to projectList being undefined and throwing an error. Earlier logic to populate projectList uses if (project) {, so using if (project && ... here keeps the checks consistent. isfsConfig() defaults project to "", which is why it is in this state. The following logic is used if apiVersion >= 6, which Cache is not: project: project ? project : undefined, // Needs to be undefined if project is an empty string. This means special handling is needed in this case


isc-bsaviano:
Hi @psteiwer, thanks for the PR! This seems like a good fix.

PR #1654, opened 2025-09-15
Lite Terminal shell integration improvements

isc-bsaviano:
- Adopts https://github.com/microsoft/vscode/issues/190253 so commands can be re-run without user confirmation
- Allow pasting of multi-line content into `READ` prompts
- Work around VS Code bug to allow re-running of multi-line commands
- Correctly capture command output in shell integration
- Fix some edge cases where the cursor and text could get out of sync

I'm opening this as a draft even though the code is ready for review because I think it should be merged just before a 3.2.0 release. This PR bumps the minimum supported VS Code version to 1.104.0 which we probably don't want to get in a 3.0.7 release if we need one.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,7 +31,7 @@
         "@types/mocha": "^7.0.2",
         "@types/node": "20.17.6",
         "@types/semver": "7.5.4",
-        "@types/vscode": "1.93.0",
+        "@types/vscode": "1.104.0",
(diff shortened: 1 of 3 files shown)


gjsjohnmurray:
Shall we merge this and prepare a 3.2.0 release for next Monday?

gjsjohnmurray (approved):
LGTM

PR #1655, opened 2025-09-16
Allow opening low-code editors from read-only file systems

isc-bsaviano:
Requires IRIS 2025.3+. I will mark this ready for review when there is a supported preview kit available.


gjsjohnmurray (approved):
happy to approve on the basis of its internal testing

PR #1656, opened 2025-09-19
Add command for showing all class members, including inherited

### LINKED ISSUE #1623: missing one place to see all elements inherited from parent classes
Type: <b>Feature Request</b>

When a class inherits from multiple other classes, there is no single view that would list all class elements (propertie, method, indices etc...) and their origin. This is very useful if project has more complex hirearchy that just a few classes. There is such feature in old legacy Studio, and it woul be very good to have it in VSCode, too.

Extension version: 3.0.4
VS Code version: Code 1.102.1 (7adae6a56e34cb64d08899664b814cf620465925, 2025-07-15T16:41:17.147Z)
OS version: Windows_NT x64 10.0.22631
Modes:

isc-bsaviano:
This PR fixes #1623. Adds a class icon button to the editor title bar when the editor contains a class. When clicked, a QuickPick appears with all members of that class, including inherited ones. Clicking on a member shows it in the editor.

all members

I initially tried to implement this as a tree view, which prompted me to make some architectural improvements. I kept those in because I think they are worth including. I consolidated three three `onDidChangeActiveTextEditor` handlers to a single one, and I added the existing tree views to the array of disposables so their resources get cleaned up properly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -225,7 +225,11 @@
         },
         {
           "command": "vscode-objectscript.showClassDocumentationPreview",
-          "when": "editorLangId == objectscript-class && activeCustomEditorId == ''"
+          "when": "false"
+        },
+        {
+          "command": "vscode-objectscript.showAllClassMembers",
+          "when": "false"
         },
         {
           "command": "vscode-objectscript.exportCurrentFile",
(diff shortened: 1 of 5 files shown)


isc-rsingh (approved):

gjsjohnmurray (approved):
Very nice. The way the symbol-class icon displays colored in the editor's action bar looks a bit weird, but that's not in our control. I'm going to open a VS Code issue about it. Approving this PR.

PR #1666, opened 2025-10-31
Prevent users from accidentally opening multiple copies of the same class or routine

isc-bsaviano:
This PR fixes an issue reported internally. A developer used `Cmd-O` to open a %Library class using the short alias. This opened a document with a different URI than if he opened it from the files explorer. This confused server-side source control and can confuse users if multiple copies are open at once. This PR fixes the issue by adding validation in our `FileSystemProvider` to reject "non-canonical" URIs. I also modified the `Open InterSystems Document...` command to be more flexible in what it accepts so opening a %Library class using its short alias is accepted and returns the "canonical" URI.


gjsjohnmurray (approved):
Looks OK but I haven't tested it.

PR #1698, opened 2026-01-20
Mapping when exporting and importing for Web Applications (such as CSPs).

GabsReDeal:
This PR allows mapping when exporting and importing for Web Applications (such as CSPs).

Currently, mapping is only implemented when exporting cls, mac, int, and inc files. This aims to allow mapping of web applications files (such as CSPs) when exporting and importing.

This aims to solve issues when working with multiple applications, but one source.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -39,7 +39,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v3
+      uses: github/codeql-action/init@v4
       with:
         languages: ${{ matrix.language }}
         # If you wish to specify custom queries, you can do so here or in a config file.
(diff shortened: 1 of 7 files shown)


isc-bsaviano:
Hi @GabsReDeal, thanks for the PR. Can you explain why you think this necessary?

GabsReDeal:
Thanks for your quick replies.

While I agree that this would he the ideal scenario, we currently have a legacy system which has some obstacles.

We are currently still on Caché and planning to move to IRIS later this year. We were not going to have an individual local instance due to the database. We currently have a large database and we are not allowed to have such a database locally for security reasons since some classes has the same data as live.

I do believe we can still overcome this by having multiple Caché databases (one for each developer) on our server instance (instead of namespaces and web applications) and then map each local instance data to that of the server.

This would still require a connection to the server to have data in the application. This could work as mostly local but the data, no?

isc-bsaviano:
If you must share a development server then you might want to use server-side source control (like Embedded Git) for now. This will allow you to edit files directly on the server like Studio and developers can share a namespace/web app. If you go with Embedded Git then it should be easy to switch to client-side editing (files on your local file system) when you can support private development environments. Client-side editing does not work well with shared development environments.

GabsReDeal:
I have little to no experience with server-side source control, and the little I have I found tedious. And the only reason we would have the "shared" development server would be to hold a database for each developer which is not on the developers' workstation.

I want to move away from a shared development environment. Without this implementation, I believe that what [I said](https://github.com/intersystems-community/vscode-objectscript/pull/1698#issuecomment-3774672112) earlier might still work for the time being as a hybrid local/server-side development. At least until we completely move to local, whenever that happens.

> I do believe we can still overcome this by having multiple Caché databases (one for each developer) on our server instance (instead of namespaces and web applications) and then map each local instance data to that of the server.

Thanks

PR #1705, opened 2026-01-28
Support deleting folders

### LINKED ISSUE #1638: Client-side editing: Deleting a folder should delete any contained documents on the server
Starting with version 3.0.0, by default this extension watches the file system for creation, change, and deletion events. Any events that happen to files that can be synced to the server will be propagated there. This works when you delete a single file, but does not if you delete an entire folder. In that case the extension should look in our index for all files that were in that folder and queue deletes for them. For this to work with web application files and "other" document types, we will probably need to add them to the index when they are created or changed, like classes and routines.

isc-klu:
# Major Changes

This PR fixes #1638. 

Previously, the plugin only supported deletion of individual files. It now deletes all files whose paths start with the affected path:
- If the affected path points to a single file, that file is deleted (unchanged behavior).
- If the affected path is a directory, all files under that directory are deleted.

To support this, `WSFolderIndex` now stores mappings only for importable files (web app files and virtual documents).

# Additional Changes

* Removed the “Launch Extension Alone” debug option, which is no longer valid since the extension now depends on InterSystems Server Manager.
* Simplified `updateIndexForDocument(uri, docs?, uris?, content?) → updateIndex(uri, content?)` because no external callers supply docs or uris.
* Simplified `removeDocumentFromIndex(uri, docs?, uris?) → removeDocumentFromIndex(uri, docs, uris)` since all callers supply the parameters.
* Simplified `isClassOrRtn(uri | str) → isClassOrRtn(str)` by requiring callers to handle the needed uri → string conversion.
* Renamed `uriIsParentOf → uriContains` because the function is checking the containment relationship (reflective, transitive closure of the parent relationship) rather than just the parent relationship.
* Loosen `uriContains` (previously, `uriIsParentOf`) to also return true when its inputs are identical documents rather than folders. 
* Deduplicated the `WSFolderIndex["documents"]` and `WSFolderIndex["uris"]` types.
* Deduplicated a repeated check used for `watcher.onDid{Change, Create, Delete}`.
* Added Prettier as an optional dev dependency to match `.vscode/settings.json` and assist future contributors with linting.

### DIFF
--- a/src/providers/LowCodeEditorProvider.ts
+++ b/src/providers/LowCodeEditorProvider.ts
@@ -9,6 +9,7 @@ import { currentFile, notIsfs, openLowCodeEditors, outputChannel } from "../util
 export class LowCodeEditorProvider implements vscode.CustomTextEditorProvider {
   private readonly _rule: string = "/ui/interop/rule-editor";
   private readonly _dtl: string = "/ui/interop/dtl-editor";
+  private readonly _bpl: string = "/ui/interop/bpl-editor";
 
   private _errorMessage(detail: string) {
     return vscode.window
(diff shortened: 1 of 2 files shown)


isc-bsaviano:
Nice first PR! In the future, please add the development branch to your fork of this repository instead of the main repo. It helps keep this repo clean.

src/utils/index.ts
```diff
@@ -801,16 +801,15 @@ export function parseClassMemberDefinition(
   };
 }
 
-/** Returns `true` if `uri1` is a parent of `uri2`. */
-export function uriIsParentOf(uri1: vscode.Uri, uri2: vscode.Uri): boolean {
-  uri1 = uri1.with({ path: !uri1.path.endsWith("/") ? `${uri1.path}/` : uri1.path });
+/** Returns `true` if `uri1` is equal to or an ancestor of `uri2`.
+ *  Non-path components (e.g., scheme, fragment, and query) must be identical.
+ */
+export function uriContains(uri1: vscode.Uri, uri2: vscode.Uri): boolean {
```
isc-bsaviano: A better name for this might be `uriStartsWith()` since it behaves like the string `startsWith()` method, just for URIs.

isc-klu: Renamed.

PR #1707, opened 2026-02-02
Move command to export Project contents to Command Palette

### LINKED ISSUE #1706: Improve the usability of the ‘Export Project Contents’ feature by making it compatible with any active server connection and available as a standalone command
Type: <b>Feature Request</b>

## Summary

The current Export Project Contents experience is confusing and fragile. It appears to work only when the active server connection is defined in the `.code-workspace` file. If the same connection is configured in a `.vscode/settings.json`, the command often refuses to run, reporting that there are “no folders in the current workspace that code can be exported to,” even though a folder is opened and the server connection is active. This has caused frustration across multiple users and community threads. [1]

Beyond the connection requirement, the command is only discoverable via the Projects explorer context menu, which is not visible unless at least one server-side virtual folder exists in the current multi-root workspace.
Making export a first-class, standalone command (Command Palette) would remove this dependency and significantly improve usability.
What is the reason for tying the visibility of the Projects Explorer to the existence of server‑side virtual folders in the workspace in the first place?

### Steps to Reproduce

1. Open a folder as your workspace root in VS Code.
2. Configure a valid InterSystems IRIS server connection in `.vscode/settings.json` and connect to a namespace. You can browse server items and compile, confirming the connection is active.
3. Open the InterSystems Projects Explorer view, right‑click a project, and choose 'Export Project Contents'.
4. Observe the error: “There are no folders in the current workspace that code can be exported to.” (despite having an open folder and an active connection).

> **Note:** If you move the identical server connection definition into the `.code-workspace` file, the export suddenly works. This dependency on where the connection is declared is unexpected and undocumented. (see post in community https://community.intersystems.com/post/how-export-project-visual-studio-code#comment-284848)

## Why This Matters

- The extension already positions export/import and client-side editing as core workflows, and many users keep their server connections in settings.json. Tying export success to the location of the connection definition makes the feature feel brittle and “random.”
- Community posts show multiple users running into the “no folders in current workspace” error even with a proper folder opened, leading to confusion and workarounds. [1][2]
- The repo history shows ongoing efforts to make Studio project export easier (e.g. #894 and related PRs/discussions), so relaxing the connection-declaration constraint and surfacing a top-level command aligns with that direction.

## Workaround (for others hitting this)

- Temporarily move the server connection definition into `.code-workspace` and retry export; then move it back if needed.
- Alternatively, use ObjectScript: Export Code from Server and/or ObjectScript: Export Documents to XML File, but those routes are less ergonomic for project-based exports.

--

[1] https://community.intersystems.com/post/how-export-project-xml-file-deploy-visual-studio-code
[2] https://community.intersystems.com/post/how-export-project-visual-studio-code

Extension version: 3.4.0
VS Code version: Code 1.108.2 (c9d77990917f3102ada88be140d28b038d1dd7c7, 2026-01-21T13:52:09.270Z)
OS version: Windows_NT x64 10.0.26200
Modes:
Remote OS version: Linux x64 5.15.167.4-microsoft-standard-WSL2
Remote OS version: Linux x64 5.15.167.4-microsoft-standard-WSL2

isc-bsaviano:
This PR fixes #1706. The existing Project content export command has been removed from the Projects Explorer and has been replaced with a new command `Export Project Contents from Server...` which can be found in the Command Palette.

### DIFF
--- a/src/commands/export.ts
+++ b/src/commands/export.ts
@@ -201,33 +201,28 @@ export async function exportAll(): Promise<any> {
     } else if (filterIsValid(filter)) {
       filters.push(`Name LIKE '%${filter}%'`);
     }
-    const files: { Name: string }[] = await api
-      .actionQuery("SELECT Name FROM %Library.RoutineMgr_StudioOpenDialog(?,?,?,?,?,?,?,?,?,?)", [
-        "*",
-        "1",
-        "1",
+    let files: vscode.QuickPickItem[] = await api
+      .actionQuery("SELECT Name FROM %Library.RoutineMgr_StudioOpenDialog('*',1,1,?,1,0,?,?,0,?)", [
         api.ns == "%SYS" ? "1" : "0",
-        "1",
-        "0",
         generated ? "1" : "0",
         filters.join(" AND "),
-        "0",
         mapped ? "1" : "0",
       ])
-      .then((data) => data.result.content);
+      .then((data) =>
+        data.result.content.map((file) => {
+          return { label: file.Name, picked: true };
+        })
+      );
     if (!files?.length) return;
-    let fileItems: vscode.QuickPickItem[] = files.map((file) => {
-      return { label: file.Name, picked: true };
-    });
-    fileItems = await vscode.window.showQuickPick(fileItems, {
+    files = await vscode.window.showQuickPick(files, {
       canPickMany: true,
       ignoreFocusOut: true,
       prompt: "Uncheck a file to exclude it. Press 'Escape' to cancel export.",
       title: "Files to Export",
     });
-    if (!fileItems?.length) return;
+    if (!files?.length) return;
     await exportList(
-      fileItems.map((file) => file.label),
+      files.map((file) => file.label),
       wsFolder.name,
       api.ns
     );


isc-klu (dismissed):

src/commands/export.ts
```diff
@@ -164,93 +164,75 @@ export async function exportList(files: string[], workspaceFolder: string, names
...
+        generated ? "1" : "0",
+        filters.join(" AND "),
+        "0",
+        mapped ? "1" : "0",
+      ])
+      .then((data) => data.result.content);
+    if (!files?.length) return;
+    let fileItems: vscode.QuickPickItem[] = files.map((file) => {
+      return { label: file.Name, picked: true };
+    });
```
isc-klu: We could replace L203–220 with the snippet below. That would remove the need for the temporary `const files` variable and simplify the logic a bit—what do you think?

```typescript
    let files: vscode.QuickPickItem[] = (
      await api.actionQuery("SELECT Name FROM %Library.RoutineMgr_StudioOpenDialog(?,?,?,?,?,?,?,?,?,?)", [
        "*",
        "1",
        "1",
        api.ns == "%SYS" ? "1" : "0",
        "1",
        "0",
        generated ? "1" : "0",
        filters.join(" AND "),
        "0",
        mapped ? "1" : "0",
      ])
    ).result.content.map(({ Name: label }) => ({ label, picked: true }));
    if (!files.length) return;
```

isc-klu (approved):

isc-klu (approved):

PR #1710, opened 2026-02-05
Remove barely used commands

isc-bsaviano:
This PR removes eight commands that are only used by a handful of users. The commands, and my reasoning, can be found below.

* `touchBar` compile and view others
  * Apple hasn't produced a Mac with a touch bar in years
* Open `subclass` and `superclass`
  * Type Hierarchy provided by Language Server is a more modern approach
* `Compile Project Contents`
  * Not sure why you'd ever need this. Doesn't line up with the purpose of the Projects Explorer (viewing and modifying projects definitions).
* Compile with specified flags
  * Flags are specified as a configuration setting. If you need to change them, change the setting.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -3,22 +3,6 @@
   "version": "0.2.0",
   "configurations": [
     {
-      "name": "Launch Extension Alone",
-      "type": "extensionHost",
-      "request": "launch",
-      "runtimeExecutable": "${execPath}",
-      "args": [
-        "--disable-extensions",
-        "--extensionDevelopmentPath=${workspaceRoot}"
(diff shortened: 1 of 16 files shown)


isc-rsingh (approved):

isc-rsingh (approved):

zvondic:
Compile with specified flags is very productive command, i don't understand why removed ? How can anybody work without this command. Please return .

isc-bsaviano:
@zvondic We already have a setting called `objectscrip.compileFlags` and various commands that compile documents without prompting you for flags. Why would you want to specify the flags every time instead of once in your settings?

PR #1730, opened 2026-03-04
Synchronize only the storage definition when compiling `*.cls` documents

### LINKED ISSUE #1729: Preserve Custom Formatting When Compiling Client‑Side Class (*.cls) Documents
Currently, custom formatting in client-side `*.cls` documents is lost during compilation. For example:

* Keywords (e.g., `classmethod`) are automatically rewritten to PascalCase (e.g., `ClassMethod`)
* Empty lines between methods are normalized to a single blank line
* Line breaks within method parameter lists are removed

It would be beneficial to decouple formatting from compilation so that user-defined formatting is preserved.

isc-klu:
Fix #1729:

* Added a new setting: `refreshClassesOnSync: boolean = false`
* Updated the import and compile logic so that when `refreshClassesOnSync` is `false`, only the storage sections of `*.cls` files are synchronized with the server version, rather than the entire file content

### DIFF
--- a/src/commands/compile.ts
+++ b/src/commands/compile.ts
@@ -118,8 +118,8 @@ export async function importFile(
     mtime < 0 ||
     (notIsfs(file.uri) &&
       vscode.workspace.getConfiguration("objectscript", file.uri).get<boolean>("overwriteServerChanges"));
-  return api
-    .putDoc(
+  try {
+    const data = await api.putDoc(
       file.name,
       {
         content,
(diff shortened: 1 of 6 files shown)


src/commands/compile.ts
```diff
@@ -318,14 +385,16 @@ export async function importAndCompile(document?: vscode.TextDocument, askFlags
...
-      .catch(() => {})
-  );
+  try {
+    if (isCompilable(file.name)) {
+      await importFile(file, true);
+      compile([file], askFlags);
+    } else {
+      await importFile(file, false);
+    }
+  } catch (_) {
```
isc-bsaviano: Line can just be `} catch {`

src/commands/compile.ts
```diff
@@ -114,78 +118,86 @@ export async function importFile(
...
-        if (!enc) {
-          choices.push("Compare");
-        }
-        choices.push("Overwrite on Server", "Pull Server Changes", "Cancel");
-        return vscode.window
-          .showErrorMessage(
-            `Failed to import '${file.name}': The version of the file on the server is newer.
+    );
+    workspaceState.update(`${file.uniqueId}:mtime`, Number(new Date(data.result.ts + "Z")));
+    if (data.result.flags !== undefined && !willCompile) {
```
isc-bsaviano: I think his check should be `!willCompile && isClass(file.name) && data.result.content.length`. You only want to run the code in this block if the file is a class and there's content to process. If not then there's nothing to update.

PR #1735, opened 2026-03-16
Disallow opening `Hidden` classes using `Open InterSystems Document...` command

isc-bsaviano:
Also adds a prompt to the QuickPick.


isc-bsaviano:
Merging based on past approvals

PR #1753, opened 2026-04-13
Remove unneeded dependencies

isc-bsaviano:
This PR removes dependencies that are no longer needed:

- `core-js`: This was only used to polyfill `flatMap` in our webpack config on older environments. Since every environment we support has `flatMap`, this can be removed.
- `vscode-cache`: This was used to cache intellisense information and session cookies. The intellisense features of this extension have been superseded by Language Server, and we no longer need to persist cookies between sessions because we log out on deactivate so this was safe to remove as well. I verified that cookies are used so sessions don't stack up, and that the logout on deactivate still happens.

### DIFF
--- a/src/utils/classDefinition.ts
+++ b/src/utils/classDefinition.ts
@@ -88,15 +88,14 @@ export class ClassDefinition {
     WHERE Name %inlist (SELECT $LISTFROMSTRING(Super, ',') FROM %Dictionary.CompiledClass WHERE Name = ?)`;
     return api
       .actionQuery(sql, [this._className])
-      .then(
-        (data) =>
-          data.result.content
-            .reduce(
-              (list: string[], el: { PrimarySuper: string }) =>
-                list.concat(el.PrimarySuper.split("~").filter((item) => item.length)),
-              []
-            )
-            .filter((name: string) => name !== this._className)
+      .then((data) =>
+        data.result.content
+          .reduce(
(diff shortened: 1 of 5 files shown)


src/utils/classDefinition.ts
```diff
@@ -58,16 +43,13 @@ export class ClassDefinition {
       if (extend.length) {
         return api.actionIndex(extend).then((data) => getMethods(data.result.content));
       }
-      return this.store("methods-" + scope, methods.filter(filterScope).filter(onlyUnique).sort());
+      return methods.filter(filterScope).filter(onlyUnique).sort();
```
isc-klu, line 46: `onlyUnique` is currently used solely to deduplicate `{ name: string }[]`. Using `<array>.filter(onlyUnique)` incurs O(n²) time complexity. I’d suggest refactoring `onlyUnique` to accept an array and return a deduplicated result (e.g., `onlyUnique(<array>)`). This would help compact the codebase and reduce the runtime cost to O(n).

isc-klu (dismissed):

src/utils/classDefinition.ts
isc-bsaviano: The code in this file is effectively dead. It's only in use when Language Server is not present. We only keep it as a backup for those users who are on platforms that Language Server does not support. I can make this change but I doubt it will impact anyone.

isc-klu (approved):

PR #1758, opened 2026-04-14
Prevent attempts to read dot-folders other than `.vscode` from reaching the server in server-side workspace folders

isc-bsaviano:
This prevents many unnecessary REST requests that occur when VS Code probes a workspace folder on load.

### DIFF
--- a/src/providers/FileSystemProvider/FileSystemProvider.ts
+++ b/src/providers/FileSystemProvider/FileSystemProvider.ts
@@ -259,6 +259,19 @@ function validateUriIsCanonical(uri: vscode.Uri): void {
   }
 }
 
+/**
+ * Throws `error` or a `vscode.FileSystemError.FileNotFound` error when `uri`
+ * is an invalid dot folder or is in an invalid dot folder. `.vscode`
+ * is allowed for any workspace folder, but other dot folders are only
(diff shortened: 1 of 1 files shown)


src/providers/FileSystemProvider/FileSystemProvider.ts
```diff
@@ -259,6 +259,19 @@ function validateUriIsCanonical(uri: vscode.Uri): void {
...
 }
 
+/**
+ * Throws `error` or a `vscode.FileSystemError.FileNotFound` error when `uri`
+ * is an invalid dot folder or is in an invalid dot folder. `.vscode`
+ * is allowed for any workspace folder, but other dot folders are only
+ * allowed in web app workspace folders. This filtering is done here to
+ * avoid spamming the server with requests that will never return meaningful data.
+ */
+function filterDotPaths(uri: vscode.Uri, error?: vscode.FileSystemError): void {
```
isc-klu: The name `filterDotPaths` doesn't match the function's behavior. _"Filter"_ typically implies returning a filtered collection, but this function validates and throws errors.

Suggested alternatives: `assertPathDotsValid`

isc-klu (changes requested):

isc-klu (approved):

PR #1765, opened 2026-04-20
Add `objectscript.insertStubContent` setting

### LINKED ISSUE #1736: Option to fully disable automatic edits to All Files
Some time ago, I opened a PR suggesting an option to completely disable the plugin’s behavior of modifying users’ file PackageName and ClassName. As far as I remember, it was closed at the time because there did not seem to be a clear use case.

Recently, I came across a scenario where this option could be genuinely useful: **Copilot Agent Mode**. In Agent Mode, the AI may try to create files from scratch and fill in their PackageName and ClassName on its own. I found that in some cases, the plugin then changes those values afterward, and Copilot may interpret the resulting file as incorrect and start editing it again. This can lead to unnecessary rework.

That behavior may also lead to additional token usage. While Copilot itself may not be billed by token usage, the plugin can also work with other tools such as Claude, where those extra edits may result in avoidable token costs.

Because of this, I think it could be helpful for the plugin to provide an option to fully disable all automatic file editing behavior, so users who prefer to let the AI handle everything can do so.

Even Copilot provides a way to disable Inline Copilot, partly to avoid extra suggestions and unnecessary token usage while using Agent Mode and simply browsing code. For a similar reason, having a full opt-out here might be useful for some workflows as well.

isc-bsaviano:
This PR fixes #1736. I would strongly prefer to avoid adding another setting but think this one may be necessary. If any of you can think of a better way to avoid the problem in the issue please let me know.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1561,6 +1561,11 @@
           "description": "If true, REST requests and responses to and from InterSystems servers will be logged to the ObjectScript Output channel. This should only be enabled when debugging a potential issue.",
           "type": "boolean",
           "default": false
+        },
+        "objectscript.insertStubContent": {
+          "description": "If true, stub content will be inserted when a blank new class or routine file is created in a client-side workspace folder.",
+          "type": "boolean",
+          "default": true
         }
       }
     },
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -807,6 +807,7 @@ function sendWsFolderTelemetryEvent(wsFolders: readonly vscode.WorkspaceFolder[]
       dockerCompose: !serverSide ? String(typeof conf.get("conn.docker-compose") == "object") : undefined,
       "config.conn.links": String(Object.keys(conf.get("conn.links", {})).length),
       "config.refreshClassesOnSync": !serverSide ? conf.get("refreshClassesOnSync") : undefined,
+      "config.insertStubContent": !serverSide ? conf.get("insertStubContent") : undefined,
     });
   });
 }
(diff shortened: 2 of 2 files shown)


isc-klu:
Let me make sure I understand the context correctly:

- We want to insert stub content into new files that are created manually by users.
- We do not want to insert stub content into new files created by AIs.
- There is no reliable way to distinguish between these two kinds of newly created files.

Given that, the best available option is to allow users to disable stub insertion entirely when their AIs may create files.
Is this understanding correct?

isc-bsaviano:
There were users who already didn't like the sub insertion. I resisted making a config setting because the stub is easy for a human to remove by undoing the insertion. The insertion is not saved. However, the AI is just not seeing that stub was inserted so it doesn't adapt or remove it. It seems to be easier to not have the stub there than to figure out some way for the AI to know about it.

isc-klu:
I believe a similar stub is already inserted when users type class and accept the first autocomplete suggestion. Are the two stubs identical? If so, removing stub insertion altogether seems more reasonable—it avoids these issues, doesn’t require introducing a new configuration, aligns better with the common behavior of language extensions, and keeps the overall cost of stub insertion low.

isc-bsaviano:
I don't see a snippet for the class header anymore (see snippets/objectscript-class.json)

isc-klu:
Sorry—that was a hallucination on my part. I also took a look at [the snippet API](https://code.visualstudio.com/docs/editing/userdefinedsnippets), and it doesn’t seem flexible enough to compute the stub. We could provide the stub via IntelliSense, but your current solution is probably the most straightforward option for now.

isc-bsaviano:
Our current solution is not ideal, but users migrating from our legacy IDE Studio are used to the stub being filled in for them. This preserves that behavior for people who see it as a time saver.

isc-klu (approved):

gjsjohnmurray (approved):
I don't have any alternative ideas

PR #1772, opened 2026-04-27
Show detailed error messages when a server-side URI is invalid

### LINKED ISSUE #1771: Saving new .MAC routine with server side connection fails with uninformative error
Are you using client-side or server-side editing?: Yes

- VS Code Version: 1.117.0
- Extension Version: 3.8.2
- OS: Windows
- Server `$ZVERSION` (if applicable): 2026.1

Steps to Reproduce:
1. Create new Test.MAC
2. Observe error

Commit 7db6cac added extra checking for URI in canonical form. As a result of this, it means that you can't create a new routine with .MAC extension, you must use .mac. If you import a .MAC routine into IRIS using the Management Portal, it accepts your routine and then displays it as .mac. If you have a client side connection with a .MAC file, it imports it as .mac.

When creating a new .MAC with server side connection, the error is a bit confusing and doesn't explain that lower case is required or what the actual problem is. There doesn't appear to be any documentation that states this either.

Should it accept the .MAC and transform it to .mac as is done in other import scenarios?

isc-bsaviano:
This PR fixes #1771
Screenshot 2026-04-27 at 4 21 06 PM

### DIFF
--- a/src/providers/FileSystemProvider/FileSystemProvider.ts
+++ b/src/providers/FileSystemProvider/FileSystemProvider.ts
@@ -247,7 +247,7 @@ export function isfsDocumentName(uri: vscode.Uri, csp?: boolean, pkg = false): s
  * and `/%Library/CHUIScreen.CLS` (extension has wrong case). This is needed to
  * prevent the user from opening multiple copies of the same document. This
  * function does not return a value; it throws a `vscode.FileSystemError.FileNotFound`
- * error, or a more descriptive custome error when `descriptiveError` is `true`,
+ * error, or a more descriptive custom error when `descriptiveError` is `true`,
  * when `uri`'s path is not in "canonical form".
  */
 function validateUriIsCanonical(uri: vscode.Uri, descriptiveError = false): void {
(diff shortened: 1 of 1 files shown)


src/providers/FileSystemProvider/FileSystemProvider.ts
```diff
@@ -247,22 +247,21 @@ export function isfsDocumentName(uri: vscode.Uri, csp?: boolean, pkg = false): s
  * and `/%Library/CHUIScreen.CLS` (extension has wrong case). This is needed to
  * prevent the user from opening multiple copies of the same document. This
  * function does not return a value; it throws a `vscode.FileSystemError.FileNotFound`
- * error when `uri`'s path is not in "canonical form".
+ * error, or a more descriptive custome error when `descriptiveError` is `true`,
```
gjsjohnmurray:
```suggestion
 * error, or a more descriptive custom error when `descriptiveError` is `true`,
```

gjsjohnmurray (changes requested):

gjsjohnmurray (approved):

PR #1788, opened 2026-05-22
Reinstate reconnect option from status bar panel of isfs-type workspace

### LINKED ISSUE #1787: Server-side (isfs) mode: Unable to trigger a reconnect from status bar
Probably since #1740, if the server referenced by and isfs-type workspace is unreachable when the workspace is initially opened (for example, if the necessary VPN connection is down), it hasn't been possible to trigger another connection attempt by clicking on the extension's status bar panel. Clicking there produces a quickpick with no options, so after fixing the network issue (e.g. starting the VPN connection) it has been necessary to reload the VS Code window, or to close and reopen the workspace.

I will submit a PR to fix this.

gjsjohnmurray:
This PR fixes #1787

### DIFF
--- a/src/commands/serverActions.ts
+++ b/src/commands/serverActions.ts
@@ -41,7 +41,7 @@ export async function serverActions(): Promise<void> {
       label: "Toggle Connection",
     });
   }
-  if (active) {
+  if (active || api.externalServer) {
     actions.push({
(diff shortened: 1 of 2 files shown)


isc-bsaviano (approved):
seems fine to me

PR #1799, opened 2026-06-23
OAuth2 VSCode

isc-klu:
This PR refactors authentication throughout the extension to use the Authorization class from @intersystems-community/intersystems-servermanager instead of directly managing username/password strings.

Key changes:

- Replace username/password properties with Authorization type in connection configurations
- Use auth.username and auth.accessToken instead of direct credential access
- Leverage auth.resolve() and auth.httpAuthorizationHeader for authentication checks
- Update REST debug panel and WebSocket terminal to use authorization methods
- Remove isUnauthenticated() helper in favor of auth.resolved() checks
- This improves security by centralizing credential handling through Server Manager's authorization abstraction and prepares the codebase for future OAuth2 support.

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -39,7 +39,6 @@ interface ConnectionSettings {
   pathPrefix: string;
   ns: string;
   authorization: Authorization;
-  password?: string;
   docker: boolean;
   dockerService?: string;
 }
@@ -314,7 +313,7 @@ export class AtelierAPI {
     headers?: any,
     options?: any
   ): Promise<any> {
-    const { active, apiVersion, host, port, password, https } = this.config;
+    const { active, apiVersion, host, port, https, authorization } = this.config;
     if (!active || !port || !host) {
       return Promise.reject();
     }
@@ -445,7 +444,7 @@ export class AtelierAPI {
         if (this.wsOrFile && !checkingConnection) {
           setTimeout(() => {
             checkConnection(
-              password ? true : false,
+              authorization.resolved(),
               typeof this.wsOrFile === "object" ? this.wsOrFile : undefined,
               true
             );
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -12,6 +12,7 @@ import {
   schemas,
   checkingConnection,
   inactiveServerIds,
+  serverManagerApi,
 } from "../extension";
 import { currentWorkspaceFolder, outputChannel, outputConsole } from "../utils";
 
@@ -19,6 +20,7 @@ const DEFAULT_API_VERSION = 1;
 const DEFAULT_SERVER_VERSION = "2016.2.0";
 import * as Atelier from "./atelier";
 import { isfsConfig } from "../utils/FileProviderUtil";
+import { IServerSpec } from "@intersystems-community/intersystems-servermanager";
 
 // Map of the authRequest promises for each username@host:port/pathPrefix target to avoid concurrency issues
 const authRequestMap = new Map<string, Promise<any>>();
@@ -37,6 +39,7 @@ interface ConnectionSettings {
   superserverPort?: number;
   pathPrefix: string;
   ns: string;
+  authMethod: "password" | "oauth2";
   username: string;
   password: string;
   docker: boolean;
@@ -59,7 +62,7 @@ export class AtelierAPI {
   }
 
   public get config(): ConnectionSettings {
-    const { serverName, active = false, https = false, pathPrefix = "", username } = this._config;
+    const { serverName, active = false, https = false, pathPrefix = "", authMethod, username } = this._config;
     const ns = this.namespace || this._config.ns;
     const wsKey = this.configName.toLowerCase();
     const host = this.externalServer ? this._config.host : workspaceState.get(wsKey + ":host", this._config.host);
@@ -83,6 +86,7 @@ export class AtelierAPI {
       superserverPort,
       pathPrefix,
       ns,
+      authMethod,
       username,
       password,
       docker,
@@ -147,12 +151,14 @@ export class AtelierAPI {
    * Manually set the connection spec for this object,
    * where `connSpec` is the return value of `getResolvedConnectionSpec()`.
    */
-  public setConnSpec(serverName: string, connSpec: any): void {
+  public setConnSpec(serverName: string, connSpec: IServerSpec): void {
     const {
       webServer: { scheme, host, port, pathPrefix = "" },
+      authMethod = "password",
       username,
       password,
     } = connSpec;
+    this._config.authMethod = authMethod;
     this._config.username = username;
     this._config.password = password;
     this._config.https = scheme == "https";
@@ -239,6 +245,7 @@ export class AtelierAPI {
     if (serverName !== "") {
       const {
         webServer: { scheme, host, port, pathPrefix = "" },
+        authMethod,
         username,
         password,
         superServer,
@@ -253,6 +260,7 @@ export class AtelierAPI {
         host,
         port,
         superserverPort: superServer?.port,
+        authMethod,
         username,
         password,
         pathPrefix,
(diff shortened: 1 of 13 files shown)


src/extension.ts
```diff
@@ -446,9 +457,11 @@ export async function checkConnection(
     panel.text = api.connInfo;
     const { serverName, host, port, pathPrefix } = api.config;
     if (serverName) {
-      panel.tooltip = new vscode.MarkdownString(`Connected to \`${host}:${port}${pathPrefix}\` as \`${username}\``);
+      panel.tooltip = new vscode.MarkdownString(
+        `Connected to \`${host}:${port}${pathPrefix}\` as \`${auth.username}\``
```
isc-bsaviano: Does this code block (and others) need to be able to handle the OAuth case where there is no username?

src/providers/LowCodeEditorProvider.ts
```diff
@@ -212,8 +212,8 @@ export class LowCodeEditorProvider implements vscode.CustomTextEditorProvider {
             webviewPanel.webview.postMessage({
               direction: "editor",
               type: "auth",
-              username: api.config.username,
-              password: api.config.password,
+              username: api.config.auth.username,
```
isc-bsaviano: What should happen here if you're connected using OAuth? The low-code editors won't be able to receive a bearer token without modification.

src/api/index.ts
```diff
@@ -385,12 +382,11 @@ export class AtelierAPI {
     let auth: Promise<any>;
     let authRequest = authRequestMap.get(mapKey);
     if (cookies.length || (method === "HEAD" && !originalPath)) {
-      auth = Promise.resolve(cookies);
-
       // Only send basic authorization if username and password specified (including blank, for unauthenticated access)
-      if (typeof username === "string" && typeof password === "string") {
-        headers["Authorization"] = `Basic ${Buffer.from(`${username}:${password}`).toString("base64")}`;
+      if (this.config.auth.resolved()) {
+        headers["Authorization"] = this.config.auth.httpAuthorizationHeader;
```
isc-bsaviano, line 387: The extension shouldn't send an Authorization header if it also sends cookies. I know this wasn;t caused by your change but it's good to clean it up now.

isc-bsaviano (changes requested):

src/extension.ts
isc-klu: Currently, when `auth` is OAuth2, the username is hardcoded to `"OAuth2User"`. Is this behavior okay?

src/providers/LowCodeEditorProvider.ts
isc-klu: Updated to fail on this case

src/extension.ts
isc-bsaviano: That's a little confusing IMO. It's unlikely, but a VS Code user could have an IRIS user with that name. If you want to hard-code a username that always means "OAuth" then I recommend you pick one that is [not a valid IRIS username](https://docs.intersystems.com/iris20262/csp/docbook/DocBook.UI.Page.cls?KEY=GSA_config_user_accounts#GSA_config_user_accounts_properties).

isc-klu: Great idea. Changed to `*OAuth2User*`

isc-bsaviano: Can we check for OAuth and show a more friendly message like `using OAuth2` instead of `as *OAuth2User*`? We also show the username in the Lite Terminal

isc-klu: If the auth is "Unauthenticated" (https://github.com/intersystems-community/intersystems-servermanager/pull/335#discussion_r3720503927), do we want to show something like "using unauthenticated connection"?

isc-bsaviano: You could say "as UnknownUser". That's the IRIS username that gets assigned to unauthenticated jobs.

isc-klu: How do you think of saying "as UnknownUser" when using OAuth2?

If we really want to say "using OAuth2", I would likely have to add one more thing to the public API (intersystems-servermanager/types/index.d.ts)

isc-bsaviano: That's inaccurate. They are logged in as a user. Won't the username be *OAuth2User* in that case?

isc-klu: For now, checking whether the username equals `*OAuth2User*` is a reliable approach. However, I'm concerned that future OAuth2 scenarios may assign real usernames to human users, which could make this assumption invalid.

isc-bsaviano: I don't think that's possible with your current implementation. When the IDP redirects back to VS Code after a successful login you get back a code, not a username. The user gets assigned a username on the IRIS side but it doesn't seem like the VS Code client will ever know what it is.

isc-klu: That's not possible at the moment. My original thought was to design the API in a way that would be future-proof, but I think I was overthinking it. If OAuth2 users are ever assigned an IRIS username, we can simply display it the same way we do for Basic Auth connections.

isc-bsaviano: Let's stick with the assumption that OAUth2 will always have the "special" username Server Manager assigns. We can always adjust later if needed.

src/providers/LowCodeEditorProvider.ts
```diff
@@ -208,13 +208,18 @@ export class LowCodeEditorProvider implements vscode.CustomTextEditorProvider {
...
+            const password = api.config.auth.password;
+            if (password) {
+              // Editor is compatible so send the credentials
+              webviewPanel.webview.postMessage({
+                direction: "editor",
+                type: "auth",
+                username: api.config.auth.username,
+                password,
+              });
+            } else {
```
isc-bsaviano, line 216: I don't think we should error in this case. I think we should show a warning message but still allow the editor to open. The user can then attempt to log in if they want. Maybe the message can be "Cannot automatically log into low-code editors if VS Code is logged in using OAuth2."

src/commands/serverActions.ts
```diff
@@ -248,7 +248,12 @@ export async function serverActions(): Promise<void> {
             if (addin) {
               sendStudioAddinTelemetryEvent(addin.label);
               let params = `Namespace=${nsEncoded}`;
-              params += `&User=${encodeURIComponent(auth.username)}`;
+              const username = auth.username;
+              if (username.includes("*")) {
```
isc-bsaviano: IMO you can still open the Add-In, just don't pass the User query parameter.

src/providers/LowCodeEditorProvider.ts
```diff
@@ -208,17 +208,24 @@ export class LowCodeEditorProvider implements vscode.CustomTextEditorProvider {
...
-                username: api.config.auth.username,
+                username,
                 password,
               });
             } else {
-              this._errorMessage("Low-code editors only support login via usernames and passwords.");
+              vscode.window.showWarningMessage(
+                `Cannot automatically log into low-code editors if VS Code is logged in using ${username.slice(1, -1)}.`
+              );
+              webviewPanel.webview.postMessage({
```
isc-bsaviano, line 217: Don't pass the `auth` message if you're not supplying username/password.

src/providers/LowCodeEditorProvider.ts
```diff
@@ -208,13 +208,23 @@ export class LowCodeEditorProvider implements vscode.CustomTextEditorProvider {
...
-              type: "auth",
-              username: api.config.username,
-              password: api.config.password,
-            });
+            const username = api.config.auth.username;
+            if (username.includes("*")) {
+              vscode.window.showWarningMessage(
+                `Cannot automatically log into low-code editors if VS Code is logged in using ${username.slice(1, -1)}.`
+              );
+              webviewPanel.webview.postMessage({
```
isc-bsaviano: Remove this. The low-code editors might not be able to handle an auth message with no username/password.

isc-bsaviano (approved):

PR #1804, opened 2026-07-08
Update `Open InterSystems Document...` command to show a flat list of documents

isc-bsaviano:
This PR implements an enhancement requested internally. A flat list of files is preferable because it helps users open their document with fewer keystrokes due to better filtering.

Current behavior:

https://github.com/user-attachments/assets/ec7ea9a2-5f02-42e9-8aad-1b57384eaf07

New behavior:

https://github.com/user-attachments/assets/7ddcdf2a-143b-4fbe-a2d9-be34c31a21d8

### DIFF
--- a/src/utils/documentPicker.ts
+++ b/src/utils/documentPicker.ts
@@ -1,6 +1,7 @@
 import * as vscode from "vscode";
 import { AtelierAPI } from "../api";
-import { cspAppsForApi, handleError } from ".";
+import { cspAppsForApi, handleError, notIsfs } from ".";
+import { isfsConfig } from "./FileProviderUtil";
 
 interface DocumentPickerItem extends vscode.QuickPickItem {
   /** The full name of this item, including its parent(s). */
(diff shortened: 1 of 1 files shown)


gjsjohnmurray:
Did you consider adding a toggle button (View as List / View as Tree) so we don't risk upsetting folk who like how it currently behaves?

isc-bsaviano:
I didn't. I'm not sure why anyone would prefer the existing behavior. I think I implemented it that way originally expecting that users would want to click through the items to select the document. Feedback from users showed that they use the input box more often. The only potential downside to this change is that loading the flat list of documents might take a lot longer than the hierarchical list. However, the QuickPick won't be locked while the list is loading so the user could type the full name or type a filter string and wait for the list to load and be filtered.

isc-klu (dismissed):

gjsjohnmurray:
@isc-bsaviano in the end it's your decision, but dropping the hierarchical approach may give rise to complaints, particularly from sites with a very large overall number of classes but a package structure that keeps lists of classes relatively short.

isc-bsaviano:
My most recent commit only shows web app files for server-side web app folders. That could significantly improve performance since IRIS wouldn't need to read the file system. While that doesn't address the UX gap for people who preferred the old way, it should ease it a little.

isc-klu (approved):

PR #1818, opened 2026-08-04
Sync server-side web app files as raw bytes

isc-bsaviano:
This PR fixes an issue reported to the WRC where web app files using non-UTF8 encoding were getting corrupted during server-side editing.

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -647,9 +647,11 @@ export class AtelierAPI {
     name: string,
     scope: vscode.Uri | string,
     mtime?: number,
-    storageOnly: boolean = false
+    storageOnly = false,
+    forceBinary = false
   ): Promise<Atelier.Response<Atelier.Document>> {
-    let params, headers;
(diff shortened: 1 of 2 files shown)


src/api/index.ts
```diff
@@ -647,9 +647,11 @@ export class AtelierAPI {
     name: string,
     scope: vscode.Uri | string,
     mtime?: number,
-    storageOnly: boolean = false
+    storageOnly = false,
+    forceBinary = false
   ): Promise<Atelier.Response<Atelier.Document>> {
-    let params, headers;
+    const params: Record<string, string> = {};
+    let headers: any;
```
isc-klu: It looks like `headers` is assigned only once, about 20 lines later, and then used immediately afterward. Would it make sense to declare headers at L672 (before `if (mtime && mtime > 0)`)? I think that would make the flow a little easier to read.

isc-klu (changes requested):

isc-klu (approved):

PR #1835, opened 2026-08-20
Skip credential resolution for inactive connections during activation

### LINKED ISSUE #1832: Cursor Remote SSH: vscode-objectscript activate() routinely hangs ~4.5 minutes after successful /api/atelier
## Summary

On **Cursor Remote SSH** (Windows client → Linux remote extension host), `intersystems-community.vscode-objectscript` activation routinely takes **~4.5–5 minutes** before the ObjectScript Explorer becomes usable. During that time the InterSystems view shows Server Manager namespaces (fast) but Explorer stays on *"There is no data provider registered that can provide view data."*

This does **not** appear to be slow IRIS/Atelier HTTP. `GET /api/atelier/` returns **200 in the same second**. The delay is almost entirely in **`activate()` resolving** (`activate resolved: ~275000ms`).

In **VS Code** (same machines / same servers), this delay does not appear to occur (or is not noticeable).

## Environment

| Item | Value |
| --- | --- |
| Editor | Cursor 1.128.0 (VS Code engine) |
| Connection | Remote SSH — client Win32, extension host Linux |
| Extension | `intersystems-community.vscode-objectscript` **3.8.4** |
| Dependency | `intersystems-community.servermanager` **3.14.1** |
| IRIS | 2024.1 (HealthShare) via HTTPS Atelier API |

Multi-root workspace with several folders using `objectscript.conn`

## Timeline (reproduced 2026-08-20)

From Cursor remote logs (`remoteexthost.log` + ObjectScript output channel):

1. **09:51:55** — `ExtensionService#_doActivateExtension intersystems-community.vscode-objectscript` (`onLanguage:objectscript-class`)
2. **09:51:55** — Server Manager activates in **~158ms** (SERVERS tree works)
3. **09:51:58** — `HEAD` + `GET https://…/api/atelier/` → **200 OK** immediately (namespaces returned)
4. **~4m 32s gap** — **no further Atelier requests** in ObjectScript HTTP log
5. **09:56:30** — `Extension activated success: intersystems-community.vscode-objectscript — **275369ms** (code loading: 73ms, activate call: 41ms, **activate resolved: 275255ms**)`
6. **09:56:30** — follow-on Atelier calls (`cspapps`, queries, Live System warning)
7. **09:56:38** — Language Server activates only *after* ObjectScript

So the Explorer “spinner forever” state is ObjectScript still inside `activate()`, not a hung network call to IRIS.

## Systemic on this remote host

Across **34** logged Cursor remote activations on the same Linux extension host:

- **Median `activate resolved` ≈ 282s**
- **28 / 34** fall in the **200–300s** band
- Occasional worse outliers (~12 min; one ~65 min)
- Rare fast cases (~41s / ~47s)

This looks like a near-fixed wait/timeout path, not random IRIS latency.

## Likely area to investigate

Connection/credential resolution appears to call:

1. `authentication.getSession('intersystems-server-credentials', …, { silent: true })`
2. If missing → `getSession(…, { createIfNone: true })`

On Remote SSH, secrets/UI live on the **client** while the extension host is **remote**. Hypothesis: `createIfNone` / SecretStorage / Accounts UI interaction stalls for ~280s under Cursor Remote even though Atelier auth for the first server already succeeded (Authorization header present on the early `/api/atelier/` call).

Suggested direction: do not block full `activate()` / explorer provider registration on interactive credential resolution for every server; defer prompts and register the data provider promptly after the first successful connection (or after silent credential lookup only).

## Secondary observation (after activate)

Language Server log then shows:

`Error making REST request HEAD : AxiosError: Request failed with status code 405`

Separate from the multi-minute activate gap, but worth noting.

## What would help

- Confirm whether `activate()` intentionally awaits `getSession({ createIfNone: true })` before registering the explorer TreeDataProvider
- Any known Cursor / Remote SSH issues with the `intersystems-server-credentials` AuthenticationProvider
- Guidance for collecting more diagnostics if needed (we can attach redacted `remoteexthost.log` + ObjectScript HTTP excerpts)

Happy to provide redacted log snippets or try a minimal workspace repro (single folder / single server) if useful.

isc-klu:
## Summary
- Fixes #1832 — `activate()` resolves credentials for every workspace folder connection, even inactive ones, before `checkConnection()` gets a chance to skip them. On slow/VPN hosts this can block activation for minutes.
- Skip credential resolution/connection checks for connections that are already inactive, in `activate()` and `onDidChangeWorkspaceFolders`.

## Test plan
- [x] `tsc --noEmit` and `eslint` pass.
- [x] Needs verification against a real slow/hanging connection (e.g. @CraigRegester's Cursor Remote SSH setup from #1832).

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -920,6 +920,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<server
   const toCheck = new Map<string, vscode.Uri>();
   vscode.workspace.workspaceFolders?.map((workspaceFolder) => {
     const uri = workspaceFolder.uri;
+    if (!new AtelierAPI(uri).active) return; // No point checking an inactive connection
     const { configName } = connectionTarget(uri);
     const conn = config("conn", configName);
 
@@ -1741,6 +1742,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<server
       const toCheck = new Map<string, vscode.Uri>();
       e.added.map((workspaceFolder) => {
         const uri = workspaceFolder.uri;
+        if (!new AtelierAPI(uri).active) return; // No point checking an inactive connection
         const { configName } = connectionTarget(uri);
         toCheck.set(configName, uri);
       });
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1745,7 +1745,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<server
       const toCheck = new Map<string, vscode.Uri>();
       e.added.map((workspaceFolder) => {
         const uri = workspaceFolder.uri;
-        if (!new AtelierAPI(uri).active) return; // No point checking an inactive connection
+        if (notIsfs(uri) && !vscode.workspace.getConfiguration("objectscript.conn", workspaceFolder).get("active"))
+          return; // Don't check inactive connections
         const { configName } = connectionTarget(uri);
         toCheck.set(configName, uri);
       });
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -923,7 +923,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<server
   const toCheck = new Map<string, vscode.Uri>();
   vscode.workspace.workspaceFolders?.map((workspaceFolder) => {
     const uri = workspaceFolder.uri;
-    if (!new AtelierAPI(uri).active) return; // No point checking an inactive connection
+    if (notIsfs(uri) && !vscode.workspace.getConfiguration("objectscript.conn", workspaceFolder).get("active")) return; // Don't check inactive connections
     const { configName } = connectionTarget(uri);
     const conn = config("conn", configName);
 
--- a/src/commands/compile.ts
+++ b/src/commands/compile.ts
@@ -283,11 +283,12 @@ function updateStorage(content: string[], storage: string[]): string[] {
   let contentString = content.join("\n");
   contentString = contentString
     // update existing Storages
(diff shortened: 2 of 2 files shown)


src/extension.ts
```diff
@@ -920,6 +920,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<server
   const toCheck = new Map<string, vscode.Uri>();
   vscode.workspace.workspaceFolders?.map((workspaceFolder) => {
     const uri = workspaceFolder.uri;
+    if (!new AtelierAPI(uri).active) return; // No point checking an inactive connection
```
isc-bsaviano: I think we should always check if it's a server-side folder. There's no way for the user to mark those connections inactive without removing the workspace folder.

isc-bsaviano:
```suggestion
    if (notIsfs(uri) && !vscode.workspace.getConfiguration("objectscript.conn", workspaceFolder).get("active")) return; // Don't check inactive connections
```
Suggesting this change based on my comment above and Craig's findings.

isc-klu: Resolved by Brett's suggestion I think

src/extension.ts
```diff
@@ -1744,6 +1745,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<server
       const toCheck = new Map<string, vscode.Uri>();
       e.added.map((workspaceFolder) => {
         const uri = workspaceFolder.uri;
+        if (!new AtelierAPI(uri).active) return; // No point checking an inactive connection
```
isc-bsaviano: Should apply the change above here too to be safe.

isc-bsaviano (approved):

PR #1840, opened 2026-08-24
Restore unauthenticated-connection support (regression from OAuth2 rewrite)

isc-klu:
## Summary
- The OAuth2 rewrite (#1799) dropped a check that skipped credential resolution for connections with no username, so servers with minimal/no security now get prompted unnecessarily. Restored (`requirePassword`).
- `resolveConnectionSpec` only cached a spec when `Authorization.resolve()` succeeded, which can never happen for an intentionally unauthenticated connection. Now caches unconditionally in that case; still skips caching on a genuine credential failure so a later call can retry.
- Fixed two crash sites (`connectFolderToServerNamespace.ts`, `addServerNamespaceToWorkspace.ts`) that assumed resolution always succeeds — `Cannot read properties of undefined (reading 'webServer')` when it doesn't.

## Test plan
- [x] `tsc --noEmit` and `eslint` pass.
- [x] Verified against a real minimal-security IRIS instance.

### DIFF
--- a/src/commands/addServerNamespaceToWorkspace.ts
+++ b/src/commands/addServerNamespaceToWorkspace.ts
@@ -47,8 +47,14 @@ async function pickNamespaceOnServer(serverName: string): Promise<string | undef
   const uri = vscode.Uri.parse(`isfs://${serverName}:%sys/`);
   await resolveConnectionSpec(serverName);
   // Prepare a displayable form of its connection spec as a hint to the user.
-  // This will never return the default value (second parameter) because we only just resolved the connection spec.
-  const connSpec = getResolvedConnectionSpec(serverName, undefined)!;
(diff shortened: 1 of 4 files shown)


isc-bsaviano:
@isc-klu Did you test that OAuth still works with this change and https://github.com/intersystems-community/intersystems-servermanager/pull/352? If so I'll approve. Code looks good.

isc-klu:
> @isc-klu Did you test that OAuth still works with this change and [intersystems-community/intersystems-servermanager#352](https://github.com/intersystems-community/intersystems-servermanager/pull/352)? If so I'll approve. Code looks good.

@isc-bsaviano Tested. OAuth2 still works with both PRs.

isc-bsaviano (approved):

PR #1844, opened 2026-08-26
Deprecate configuring a server host/port/credentials in `objectscript.conn`

isc-bsaviano:
This PR adds a deprecation message that will be shown whenever a user configures a server directly in `objectscript.conn`. This legacy mechanism is leftover from before Server Manager existed and shouldn't be used anymore. I only added the message to the `port` property so the user won't see four messages if they have all four properties defined, and since `port` has non default it must always be set to be usable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1196,6 +1196,7 @@
             "port": {
               "type": "integer",
               "description": "TCP port number the web server listens on.",
+              "deprecationMessage": "It is recommended to define a server connection in `#intersystems.servers#` instead of here.",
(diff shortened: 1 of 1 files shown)


gjsjohnmurray (approved):
Expect to hear grumbles from some old-school users though.

isc-bsaviano:
There are no plans to remove this feature. The message reflects the reality that we don't test for this edge case.

isc-klu (approved):

PR #1846, opened 2026-08-26
`serverForUri` should preserve passwords stored in `settings.json["objectscript.conn"]`

### LINKED ISSUE #415: InterSystems Server Credentials request
What happens ?
what I'm now getting this annoying message?
How to disable it?
I have objectscript.conn configured, and it's enough for me, was working fine until I started to get this message for no reason

Image

isc-klu:
Fix https://github.com/intersystems/language-server/issues/415, which I believe is a regression introduced by 5d4083016933f48bbe5a5bc08b38e12fdc86a2cc

This PR might be related to https://github.com/intersystems-community/intersystems-servermanager/issues/353

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -2014,24 +2014,22 @@ function serverForUri(uri: vscode.Uri): serverManager.ServerForUri {
   // which will require explicit user consent to divulge the password to the requesting extension.
   const { serverName, active, host, https, port, superserverPort, pathPrefix, auth, ns, apiVersion, serverVersion } =
     api.config;
-  // The auth is resolved if and only if a password is provided by settings.json["objectscript.conn"]
-  if (! auth.resolved()) {
-    const password: string | undefined =
-      serverName &&
-      vscode.workspace
-        .getConfiguration(
-          `intersystems.servers.${serverName.toLowerCase()}`,
-          // objectscript(xml):// URIs are not in any workspace folder,
-          // so make sure we resolve the server definition with the proper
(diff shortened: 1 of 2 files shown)


gjsjohnmurray (approved):

src/extension.ts
```diff
@@ -2015,8 +2015,8 @@ function serverForUri(uri: vscode.Uri): serverManager.ServerForUri | undefined {
   const { serverName, active, host, https, port, superserverPort, pathPrefix, auth, ns, apiVersion, serverVersion } =
     api.config;
   auth.clear() as void;
-  const password: string | undefined =
-    serverName &&
+  const password =
+    config("conn", configName).password ||
```
isc-bsaviano, line 2019: Any reason why you used `config()` instead of `vscode.workspace.getConfiguration()`

isc-klu: No, I notice that `config()` is used at many places and it seems to contain some preprocessing that I don't understand, so I use it to be safe.

isc-klu: After checking config()'s source, I think we do want to use `config()` because some configured passwords are not stored in the normal way:

https://github.com/intersystems-community/vscode-objectscript/blob/f86793f9339cba45494c83e1afe0c269ca433e65/src/extension.ts#L195

isc-bsaviano: That's gross! Have to keep it for backwards compatibility though...

isc-bsaviano (approved):

PR #1859, opened 2026-09-04
Podman support

isc-klu:
I implemented this so I could test docker-compose connection scenarios without installing docker myself. Not sure there's broader interest in podman support — feedback welcome.

## Summary
- `composeCommand()` falls back to `podman-compose` when neither `docker compose` nor `docker-compose` is present.
- Added `portFromPodmanCompose`, used only for podman: its `port` output is a bare port number (not `host:port`), and it lacks `ps --services --filter status=running`, so it needs its own resolution via `ps --format json`.
- Docker-compose path is untouched; diff is purely additive aside from one dispatch branch.

## Test plan
- Verified against a real IRIS container under `podman-compose`: resolves both ports, reports failures correctly.
- `tsc`/`eslint` clean.


src/utils/index.ts
```diff
@@ -1,5 +1,6 @@
 import path = require("path");
 import { exec } from "child_process";
+import { promisify } from "util";
```
isc-bsaviano: do we really need `promisify`?

src/utils/index.ts
```diff
@@ -468,11 +469,80 @@ async function composeCommand(cwd?: string): Promise<string> {
         // 'docker compose' is not present, so default to 'docker-compose'
         cmd = "docker-compose";
       }
-      resolve(cmd);
+      exec(`${cmd} version`, { cwd }, (error) => {
+        if (error) {
+          // Neither Docker Compose variant is present; fall back to Podman Compose
+          cmd = "podman-compose";
```
isc-bsaviano: Doesn't podman also support `podman compose` syntax? https://docs.podman.io/en/latest/markdown/podman-compose.1.html

isc-klu: Good catch — added it.

src/utils/index.ts
isc-klu: No — dropped it.

isc-bsaviano (approved):

PR #1864, opened 2026-09-08
Fix connection check prompting/disabling on a merely-stale session

### LINKED ISSUE #1861: Connection check runs on every editor change and turns an expired session into a password prompt; dismissing it disables a working connection
Are you using client-side or server-side editing?: client-side

- VS Code Version: 1.136.0 (520fb30b2d3d324b4cb2342f6e88e2cd93751de1, x64)
- Extension Version: 3.8.5 (Server Manager 3.14.2)
- OS: Windows Server 2022 Standard, 10.0.20348
- Server `$ZVERSION` (if applicable): IRIS for Windows (x86-64) 2026.3.0AI (Build 136U) Tue Aug 18 2026 09:45:27 EDT
- Web server: IIS 10 with the ISC Web Gateway (relevant — the 401 below comes from there)

Steps to Reproduce:

1. Open a client-side workspace folder whose `objectscript.conn` points at a named server in
   `intersystems.servers`. No `password` in settings anywhere; enter the credential once when
   asked, so it is cached in `workspaceState`. Confirm the connection works.
2. Let the server's `/api/atelier` session expire. Waiting past the web application's session
   timeout is enough; setting that timeout low makes it deterministic (5 seconds here).
3. Change the active editor to **any** file in the workspace folder. A Markdown file in an
   unrelated folder is enough — it does not have to be an ObjectScript document.
4. `Not Authorized. Enter password to connect as user X to Y` appears.
5. Leave the prompt open and run Import and Compile on a routine. It succeeds.
6. Press `Escape`. `objectscript.conn.active` is set to `false` in the workspace settings file
   and the connection is gone until it is re-enabled by hand.

#### Relation to #1858 and #1860

Same symptom, different mechanism — filed separately on purpose.

#1858 and the draft fix in #1860 concern a plaintext password in `objectscript.conn` being
discarded during spec resolution for docker-compose connections, so `auth.resolved()` returns
`false` and the prompt path is taken from the start. #1860 is a one-line change
(`auth: serverForUri.auth ?? new BasicAuthorization()`); it does not touch the connection
check, the `onDidChangeActiveTextEditor` handler, the 401 handling, or the `workspaceState`
password cache.

This report is the opposite case: no password is stored anywhere, the connection uses a named
server, `auth.resolved()` is `true`, and the credential is resolved and used **successfully**
before being discarded after a 401.

#### Why the document is irrelevant

The connection check is invoked from the `onDidChangeActiveTextEditor` handler for every
editor, with no filter on `languageId` or URI scheme — while the statement immediately after
it in the same handler does filter on the class language id for `openClassContracted`. The
check resolves its target from the workspace folder of the active document, so every file in
the folder maps to the same connection; a document outside any folder still falls back to the
first workspace folder.

The request that fails is `GET /api/atelier/` — server metadata, nothing document-specific.
Its result is used to set the status bar text and tooltip. A status-bar refresh should not be
able to reach a credential prompt.

#### Measured

With `objectscript.outputRESTTraffic` enabled; values are masked by the extension itself.

The failing check — a session cookie was sent, and because a cookie was present no credentials
were attached:

```
+- REQUEST - 12:40:49 ---
GET https://host:443/api/atelier/
COOKIE: <value>
ACCEPT: application/json
CACHE-CONTROL: no-cache          <- no AUTHORIZATION header
+- RESPONSE - 12:40:50 ---
401 Unauthorized
WWW-AUTHENTICATE: Basic
```

The credentials were never the problem. Five seconds later, after the prompt was answered:

```
+- REQUEST - 12:40:54 ---
HEAD https://host:443/api/atelier/
COOKIE: <value>
AUTHORIZATION: <value>
+- RESPONSE - 12:40:54 ---
200 OK
SET-COOKIE: <value>
```

#### The prompt is raised for a condition that has already been repaired

Step 5 above, from the same log. While the dialog sat open and untouched:

```
PUT  .../v8/NS/doc/SOME.mac?ignoreConflict=1
COOKIE: <value>
-> 401 Unauthorized

HEAD .../api/atelier/
COOKIE: <value>
AUTHORIZATION: <value>
-> 200 OK   SET-COOKIE: <value>

POST .../work -> 200 · GET .../work/37993851 -> 200 · GET .../doc/SOME.mac -> 200
```

The ordinary request path recovers on its own: the 401 drops the cookie for that target, the
next request therefore has none, credentials are attached, a new session is issued, work
continues.

The connection check never gets there. Its own 401 arrives while `checkingConnection` is
`true`, and the single retry is guarded on that flag, so it is skipped. The check is then
called with `reset = auth.resolved()`, and the reset **deletes the cached password** from
`workspaceState` — discarding the credential the extension is at that moment using
successfully for other requests. The next editor change asks for it again.

Across the whole session log there were 4 × 401 against several hundred successful requests,
and the last entry while the dialog was still open was a `200 OK`.

#### Dismissing the dialog disables a working connection

`Escape` adds the server to `inactiveServerIds` and writes `objectscript.conn.active: false`
into the workspace settings file. The dialog does announce this ("If no password is entered
the connection will be disabled"), but at that moment the connection is not broken — it is
demonstrably serving requests.

#### Environment note, offered up front

The `/api/atelier` web application here had its session timeout set to 5 seconds, a deliberate
change to reduce CSP license-unit pressure from automated clients. That is why this fires on
nearly every editor change in this installation rather than occasionally. It is not the
defect, only its amplifier: any expired session, server restart or network interruption
produces the same sequence at a normal timeout, just less often — and the recovery the
ordinary request path performs shows the extension already knows how to handle it.

#### Suggestions

1. Scope the editor-change connection check to documents the extension owns — its own language
   ids and the `isfs` / `objectscript` schemes — the way the following statement in the same
   handler already does.
2. Do not let a status-bar refresh escalate to a credential prompt. Setting the panel to a
   disconnected state and letting the next real request re-authenticate would be enough.
3. On a 401 where a cookie was sent and no `Authorization` header was, retry with credentials
   before prompting — the same recovery the ordinary request path already performs. At minimum,
   do not clear the cached credential on a path where it has not been shown to be wrong.
4. Do not disable the connection when the dialog is dismissed, at least not while the
   credentials are known and other requests are succeeding.

isc-klu:
Fixes #1861

**Root cause**: `checkConnection`'s probe `request`s always passed `retryAfter401=false` to `AtelierAPI`. This is likely intended to prevent a recursive loop back into `checkConnection`. But it also disables `request`'s unrelated single-retry self-heal, which is what should fix a merely-expired session without user involvement.

**Fix**: Dropped the flag and moved all recursion-prevention logic to `checkConnection` (now `ensureConnection`) itself with a `configName`-scoped `Set` (`ensuringConnection`) instead of the unscoped boolean.

Cleanup that fell out of this:
* Renamed `checkConnection` to `ensureConnection` — it's not just a check, it has side effects (prompts, workspace state, etc.).
* Renamed its `inActivate` param to `withTimeout`, matching what the boolean actually controls.
* Tightened `request()`'s signature: `method: string` -> literal union; `options: any` -> a concrete type.
* Dropped a couple of redundant `checkingConnection` resets already covered by the trailing `.finally(() => { ... })`.


isc-bsaviano:
Nice, looks good. Will wait to approve until issue reporter responds.

isc-bsaviano (approved):
Upvoting so this can be merged after final confirmation.

Issue #1862, opened 2026-09-07
Lite Terminal: pasted input that ends with a newline is not syntax colored


globalerp-mhess:
Are you using client-side or server-side editing?: client-side

- VS Code Version: 1.136.0
- Extension Version: 3.8.6-dev.1860 (also reproduces on `master`, 6b74b6f)
- OS: Windows Server 2022 Standard, 10.0.20348
- Server `$ZVERSION` (if applicable): IRIS for Windows (x86-64) 2026.3.0AI (Build 136U) Tue Aug 18 2026 09:45:27 EDT

Steps to Reproduce:

1. Open a Lite Terminal.
2. Copy these two lines out of a Markdown code block, so the copied text ends with a newline:

```
write "Hello "
write "world!"

```

3. Paste them into the terminal.

The input runs immediately and stays uncolored. Pasting the same text *without* the
trailing newline and pressing Enter colors it as expected, so the only difference is
whether the pasted text carries its own carriage return. That is what makes it a common
case rather than a corner one: copying a code block out of a Markdown file sometimes appends one.

The cause is visible in the source, so I skipped the WebSocket trace. In the `default`
branch of `handleInput` in `src/commands/webSocketTerminal.ts`, the `{"type": "color"}`
request sits in the `else if` to `if (submit)`. Input that arrives with its own carriage
return is therefore sent for evaluation without ever making the color round-trip.

isc-bsaviano:
This is working as designed. Coloring is skipped in this case because it doesn't help the user write the command; the command already been sent for execution. I can see how this may be jarring though. @isc-klu if this is trivial to address then let's address it.

globalerp-mhess:
Thanks. That makes sense for the writing phase. 
What still bites is the reading afterwards: 
the submitted line stays uncolored in the scrollback for good, right next to typed commands or pasted commands without newline that are colored, so the transcript is inconsistent for anything you scroll back to. 

Where I ran into it: 
I'm putting together an in-house training in which I want to demonstrate ObjectScript by pasting prepared snippets into the Lite Terminal. 
The room will be reading the screen, and the pasted lines come out uncolored and will be hard to read. 

I have a patch running here if it's useful: 
the paste path asks for the colored input, writes the reply, and submits afterwards. Same socket, same order, so the reply always lands before any output. 
The honest cost is one round-trip before the command is dispatched, plus a fallback timer so a missing `color` reply can't strand the input. 

[...]

PR #1867, opened 2026-09-10
Lite Terminal: don't repaint a pasted block taller than the viewport

globalerp-mhess:
Follow-up to #1866 and #1862, which fixed the coloring itself. One case still misdraws: a paste
taller than the terminal window.

Coloring a submitted paste repaints the input in place — save the cursor, move up the number of rows
computed for the input, erase, redraw, restore. That upward move is unfulfillable once the input is
taller than the window, because `\x1b[nA` stops at the top row. The redraw then starts as many rows
too low as the input exceeds the viewport, which leaves the top of the original block on screen and
puts the command's output inside the second copy.

Measured with a 31-line paste of `set a01=1` … `set a30=30`: in a 49-row window the handler moves up
30 rows and the result is correct; in a 20-row window it computes the same 30 and lands 11 rows off —
exactly the overflow. Nothing in the class can notice, since `setDimensions` receives
`dimensions.rows` but only `columns` is kept. On the commit before #1866 the same paste renders
correctly, so it is the repaint and not the raw echo: a downward correction clamping at the bottom
row happens to land where it wants to be, an upward one does not.

The change: input that arrives with its own carriage return is submitted straight away, so its echo
is never corrected by a later keystroke. Hold that echo back and let the `color` handler write the
colored text in its place. A plain write at the position the raw text would have occupied has no
arithmetic in it, so the height of the block stops mattering. The 100 ms window and the
`_state != "prompt"` guard from #1866 are unchanged; if no coloring arrives in time the raw text is
written instead — uncolored as before, but never invisible. Typed input keeps the existing repaint,
which is correct for it, since every keystroke colors a block already on screen.

Verified against IRIS 2026.3.0AI: the 31-line paste is colored and correct in a 20-row window, where
it was misdrawn before, and unchanged in a tall window; typed multi-line input with Enter behaves as
before.

Edit for clarification:

**With #1866 in place — 31-line paste into a 20-row window.** The block is drawn twice and the
command's output lands inside the second copy:
grafik

**Same paste, same window, with this change:**
grafik


isc-bsaviano:
Thanks for the PR @globalerp-mhess!

@isc-klu This seems fine to me but I want your review since it's related to your change.

src/commands/webSocketTerminal.ts
```diff
@@ -297,6 +300,13 @@ class WebSocketTerminal implements vscode.Pseudoterminal {
           case "color": {
             // Outdated: the input is no longer on screen
             if (this._state != "prompt") break;
```
isc-klu, line 302: @isc-bsaviano I am not sure if we still need to check `this._state` but we can leave it as a future task for ourselves.

isc-klu (approved):

isc-bsaviano (approved):

isc-bsaviano:
Thanks again @globalerp-mhess! You'll get a shoutout in the CHANGELOG when we prepare the next release.

Issue #1863, opened 2026-09-07
Pre-release unit tests


isc-bsaviano:
Our current unit tests are useless. Let's remove them and replace them with a more comprehensive suite.

- Only runs on PRs with a source branch whose name starts with "prepare-". This will allow us to run more costly tests but still catch issues before a release.
- Test should spin up an iris-community container (use latest-cd tag). It will need to set a password for the pre-defined system usernames at startup time.
- Test should have a multi-root workspace that attempts to connect to the server using a different mechanism (client-side using ObjectScript.conn, client-side using docker-compose port resolution, server-side using an InterSystems.servers entry that has password in settings.json).
- Once connection is established on all folders, attempt some common actions. For example, try to save a file and see if it syncs properly.

Let's try to get this done before we do another minor release (not patch).

PR #1869, opened 2026-09-10
Fix the test runner so `npm test` can launch VS Code again

isc-klu:
`npm test` has not been able to launch VS Code for a while. These are the runner-only fixes needed before replacing the suite itself (#1863).

- `@vscode/test-electron` 2.x looks for `Contents/MacOS/Electron`, which VS Code 1.110+ renamed, so every macOS run failed with `spawn ... ENOENT`. Bumped to 3.x; its typings need `skipLibCheck` with our pinned `@types/node`.
- Dropped the `-n` launch arg, which Electron rejects (`bad option: -n`) now that the runner spawns it directly.
- Unset `ELECTRON_RUN_AS_NODE`, inherited when running from an extension-spawned terminal. Otherwise the downloaded VS Code runs as plain Node and tries to execute the workspace file as a script.
- Fresh `--user-data-dir` per run so cached state can't leak between runs.
- The mocha runner returned a promise that settled before the tests ran, so failures never failed `npm test`.


isc-klu:
Folded into #1870.

PR #1870, opened 2026-09-10
Integration tests against an IRIS container for release-preparation PRs

### LINKED ISSUE #1863: Pre-release unit tests

isc-klu:
Closes #1863.

## Reading order

1. `test-fixtures/README.md` — the design: cases, checks, containers, how to run locally.
2. `src/test/cases.ts` — that matrix as data; `src/test/suite/extension.test.ts` — the checks.
3. `src/test/runTest.ts` and `.github/workflows/prepare-release.yml` — how they get run.

Companion of intersystems-community/intersystems-servermanager#359; both repos run the same matrix, each installing the other extension's Marketplace release.

## What changes

- New `prepare-release.yml` workflow runs the suite against IRIS containers on `prepare-*` PRs and manual dispatch.
- The placeholder `npm test` (a single `assert.ok`) is replaced, and the step leaves `main.yml` and `prerelease.yml`.
- `@vscode/test-electron` 2.x → 3.x: 2.x looks for `Contents/MacOS/Electron`, which current VS Code renamed, so `npm test` could not launch at all.

## Decisions worth knowing

- **Podman only.** The workflow removes the runner's Docker Compose (the extension tries it first), and `runTest.ts` pins `PODMAN_COMPOSE_PROVIDER=podman-compose` (Docker Desktop's plugins hijack `podman compose`).
- **Isolation.** Fresh `--user-data-dir` per case; the suite asserts it is running the build under test, not a Marketplace copy.
- **Lapsed sessions.** 10 s `/api/atelier` timeout; every check reruns as the first request after it lapses. Found #1871, fixed by #1872 (included).
- **Runner.** Dropped the `-n` arg Electron rejects, unset an inherited `ELECTRON_RUN_AS_NODE`, and made the mocha promise settle when the tests finish.


isc-bsaviano (approved):
This looks awesome KC! I'm going to approve and merge so the tests get run for the next release I prepare. A good next set of cases would be to test this extension's exported API.

PR #1873, opened 2026-09-16
Prepare 3.8.6 release

isc-bsaviano:

PR #1875, opened 2026-09-17
Fix the docker-compose integration cases on CI

isc-klu:
Red on #1873. Two one-liners: the compose file path is relative to the workspace folder and broke when the client folder moved under `.generated/`; and ubuntu-latest installs the Docker Compose plugin as a binary, not the `docker-compose-plugin` package, so the extension kept picking `docker compose` over Podman.

Also quiets the launched VS Code (`chat.disableAIFeatures`, `--disable-gpu`): Copilot Chat and GPU probing were ~90% of the log, hiding the failures.


isc-bsaviano:
@isc-klu Do you want to make these changes as part of my existing PR so they can be tested right away?

isc-klu:
> @isc-klu Do you want to make these changes as part of my existing PR so they can be tested right away?

That's a great idea. What's the best way to do that?

isc-bsaviano:
I have the PR set to allow contributions from maintainers. Maybe you need to git clone my fork and then push to the prepare branch?

isc-klu:
Close as folded into the prepare- PR

